PR #3, opened 2023-05-05
[STK-510] Next js and basic setup

wixzi:
Next js and basic setup


berteotti:
I've some questions about the initial setup.

- Should we use `/appDir`?
- Do we need the `ui` package? If we are creating other packages than `app` (like landing page), I can see a big reason for it.
- The same think for `ui` applies to `utils`.

Otherwise, good job Velu!

Diogomartf:
> I've some questions about the initial setup.
> 
> * Should we use nextjs `/app`?
> * Do we need the `ui` package? If we are creating other packages than `app` (like landing page), I can see a big reason for it.
> * The same think for `ui` applies to `utils`.
> 
> Otherwise, good job Velu!

Let's use the app router and the new file system.

https://nextjs.org/docs/app/building-your-application/routing

image

packages/app/next.config.js
```diff
@@ -0,0 +1,6 @@
+/** @type {import('next').NextConfig} */
+const nextConfig = {
+  transpilePackages: ["@stackly/ui"],
```
Diogomartf, line 3: Not sure if we need the UI outside of app at this point.

The only potential use for the UI lib is if we build the landing page on other repo and want to use some components.

packages/app/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "start": "next start",
+    "lint": "next lint",
+    "export": "next export",
+    "typecheck": "tsc --noEmit",
+    "clean": "rm -rf .vercel && rm -rf .turbo && rm -rf out && rm -rf .next"
+  },
+  "dependencies": {
+    "@stackly/ui": "*",
+    "@stackly/utils": "*",
+    "next": "canary",
```
Diogomartf, line 19: I think we should specify a version, like the latest 13.4 

https://nextjs.org/blog/next-13-4

Diogomartf:
> * Should we use nextjs `/app`?

I do think so, we should use the new app router. I've been using it to rebuild my website and has been great. And I think it will be even better as a team because there are a lot of good conventions.

> * Do we need the `ui` package? If we are creating other packages than `app` (like landing page), I can see a big reason for it.
> * The same think for `ui` applies to `utils`.

The question here is: Can we forecast the use the `ui` lib in other repos? If not we can move the `ui` to inside `app`.
Same with utils, will we share utils across repos or packages?

packages/app/next.config.js
wixzi: Moved inside the app

packages/app/package.json
wixzi: Added proper version

berteotti (approved):
Looks awesome, good job Velu!

PR #4, opened 2023-05-06
[STK-61] Add favicons and touch icons

wixzi:
- This PR adds favicons and app touch icons
- This should be merged after #3 so this has only icon change commits


packages/app/app/layout.tsx
```diff
@@ -0,0 +1,47 @@
...
+        />
+        <link rel="manifest" href="/assets/favicons/site.webmanifest" />
+        <link
+          rel="mask-icon"
+          href="/assets/favicons/safari-pinned-tab.svg"
+          color="#F8F8F7"
+        />
+        <meta name="msapplication-TileColor" content="#F8F8F7" />
+        <meta name="theme-color" content="#A2E771" />
+      </head>
```
Diogomartf, line 43: with new app router we don't need that to set up the icons.

Just add the favicon.ico to the root of the project

packages/app/public/assets/favicons/android-chrome-192x192.png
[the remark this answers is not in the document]
wixzi: You mean this?
https://www.figma.com/file/XsJ1Cwv46rhwVOWzlAdhhH/Stackly-V1.0?type=design&node-id=723-18261&t=dPjCUB4yFCHNXu20-4

Then how its visible in light mode?

packages/app/app/layout.tsx
wixzi: That's why I added the favicon.icon to the public but you need to code the below files for apple and icon specific to devices based on the size for device specific responses.

`Add an icon.(js|ts|tsx) or apple-icon.(js|ts|tsx) file that default exports a function that returns Blob | ArrayBuffer | TypedArray | DataView | ReadableStream | Response.`

 Or use what I did above. I thought the above was easy

packages/app/public/assets/favicons/android-chrome-192x192.png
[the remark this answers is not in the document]
wixzi: I think this can be merged now and later change if @akashtyagi027 wants to change.

packages/app/app/layout.tsx
wixzi: That is what I understand from the doc. If I am wrong please advice. I can change.

Diogomartf: You can put the favicon.icon on the root folder (not in public) and it will work.

Diogomartf: With app router you just need to add the correct filenames to work:

Just add this files to the root:
- favicon.ico
- icon.(jpg|jpeg|png|svg)
- apple-icon.(jpg|jpeg|png|svg)

And you can remove this `head`.

Docs:
image

packages/app/public/assets/favicons/android-chrome-192x192.png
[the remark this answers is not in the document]
wixzi: Updated and removed all other files

berteotti (dismissed):
Looks good to me! Just a question, should we have other icons for different devices?

Diogomartf (approved):

PR #6, opened 2023-05-06
[STK-62] Stackly header and logo

wixzi:
## Description
- Added Navbar
- Mobile menu
- Mobile hamburger button and close
- Mobile menu backdrop
- Placeholder fields for wallet connect and user preview


berteotti:
Can you update .gitignore so there aren't build files on our repo?

berteotti:
There is a mobile version on figma. Stackly header should adjust itself if is rendered on mobile or desktop.

Diogomartf:
Some things are missing on the navbar.

I let some prints just for reference:

image

image


On mobile it doesn't need to ship with the correct buttons, but it has to have the structure to handle that positions.

wixzi:
> Some things are missing on the navbar.
> 
> I let some prints just for reference:
> 
> image image
> On mobile it doesn't need to ship with the correct buttons, but it has to have the structure to handle that positions.
> 
> Make sure to also use the correct icons for the hamburguer and close icon (they also have bg color).
> 
> You can merge with develop to get the tailwind theme.

Thanks for checking

packages/app/components/Header/HamburgerButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+  toggle: () => void;
+}
+
+export default function HamburgerButton({
+  open,
+  toggle,
+}: HamburgerButtonProps) {
+  return (
+    <button
+      className="space-y-1.5 md:invisible px-[10px] py-[10px] rounded-md  bg-[#F3F4F2] active:bg-[#ECEDEB]"
```
berteotti, line 16: Maybe we should define a `min-width` so it doesn't collapse on mobile.

packages/app/package.json
```diff
@@ -25,6 +25,7 @@
     "eslint": "^8.11.0",
     "eslint-config-next": "canary",
     "postcss": "^8.4.23",
+    "sharp": "^0.32.1",
```
berteotti, line 28: Why are we adding this lib?

packages/app/components/Header/MobileMenuList.tsx
```diff
@@ -0,0 +1,38 @@
...
+export default function MobileMenuList({ open, toggle }: MobileMenuListProps) {
+  if (!open) return null;
+
+  return (
+    <>
+      <div
+        className="fixed top-[72px] bottom-0 left-0 right-0 bg-gray-alpha-75"
+        onClick={toggle}
+      ></div>
+      <div className="w-full bg-surface-25 rounded-lg absolute top-[72px] left-0 border-b border-solid border-surface-75 gap-2 py-2 px-6">
```
berteotti, line 19: Wondering if we should have `72px` or just `72` somewhere as I'm seeing this value in various files/elements

packages/app/components/Header/HamburgerButton.tsx
berteotti: Also should we had more border radius to be closer to the design?

berteotti (changes requested):
Good job Velu, header looks nice.  Just wondering if we should have the mobile logic all in the same component, using components like the placeholders and your stacks link in just 1 place.

packages/app/components/Header/Header.tsx
```diff
@@ -0,0 +1,47 @@
+"use client";
```
Diogomartf, line 1: why do we need the "use client" here?

packages/app/components/Header/Header.tsx
```diff
@@ -0,0 +1,47 @@
...
+import Logo from "../../public/assets/images/logo.svg";
+import Stackly from "../../public/assets/images/stackly.svg";
+import Blocks from "../../public/assets/images/blocks.svg";
+
+import Link from "next/link";
+import MobileMenu from "./MobileMenu";
+
+export default function Header() {
+  return (
+    <header className="h-[72px] sticky top-0 z-50 flex flex-col w-full bg-surface-25 border-b border-solid border-b-surface-75">
```
Diogomartf, line 12: I don't think we want the nav `sticky`, at least by default.  Also z-index is to high, let start with `z-10`.

```suggestion
    <header className="h-[72px] top-0 z-10 flex flex-col w-full bg-surface-25 border-b border-solid border-b-surface-75">
```

packages/app/components/Header/Header.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <nav className="h-full flex items-center w-full max-w-screen-xl px-4 mx-auto">
+        <div>
+          <Link href="/" title="Stackly Home" className="flex items-center">
+            <Image aria-label="Stackly logo" src={Logo} alt="Stackly Logo" />
+            <Image
+              aria-label="Stackly"
+              src={Stackly}
+              alt="Stackly Logo"
+              className="ml-[12px] hidden md:block"
```
Diogomartf, line 21: Let's use the default scale when possible. 

`ml-3 == 12px`

```suggestion
              className="ml-3 hidden md:block"
```

Diogomartf: This is extension makes using tailwind really easy. You can even hover on classes to get more info like the equivalent px.

**Tailwind IntelliSense**
https://github.com/tailwindlabs/tailwindcss-intellisense

**Headwind**
Keeps the classnames sorted the same.
https://github.com/heybourn/headwind

packages/app/components/Header/Header.tsx
```diff
@@ -0,0 +1,47 @@
...
+            />
+          </Link>
+        </div>
+        <div className="border-r border-solid border-b-gray-100 h-[31px] ml-[20px] md:ml-[40px]"></div>
+        <div className="justify-end items-center gap-4 w-full hidden md:flex">
+          <Link href="#" className="flex py-3 items-center text-med-em">
+            <Image
+              alt="your stacks"
+              src={Blocks}
+              className="w-[18px] h-[18px]"
```
Diogomartf, line 31: It's better to use Image `width` and `height` props.

```suggestion
            <Image
              alt="your stacks"
              src={Blocks}
              width={18}
              height={18}
```

packages/app/components/Header/MobileMenuList.tsx
```diff
@@ -0,0 +1,38 @@
...
+  toggle: () => void;
+}
+
+export default function MobileMenuList({ open, toggle }: MobileMenuListProps) {
+  if (!open) return null;
+
+  return (
+    <>
+      <div
+        className="fixed top-[72px] bottom-0 left-0 right-0 bg-gray-alpha-75"
```
Diogomartf, line 16: I don't think we want the top nav fixed when menu not open.

packages/app/app/layout.tsx
```diff
@@ -14,7 +15,10 @@ interface RootLayoutProps {
 export default function RootLayout({ children }: RootLayoutProps) {
   return (
     <html lang="en">
-      <body className="font-sans bg-surface-25">{children}</body>
+      <body className="font-sans bg-surface-25">
+        <Header />
```
Diogomartf, line 19: Let's call it Navbar

```suggestion
        <Navbar />
```

packages/app/components/Header/Header.tsx
```diff
@@ -0,0 +1,47 @@
...
+        <div>
+          <Link href="/" title="Stackly Home" className="flex items-center">
+            <Image aria-label="Stackly logo" src={Logo} alt="Stackly Logo" />
+            <Image
+              aria-label="Stackly"
+              src={Stackly}
+              alt="Stackly Logo"
+              className="ml-[12px] hidden md:block"
+            />
+          </Link>
```
Diogomartf, line 23: I would extract this into `Logo` component

packages/app/components/Header/MobileMenuList.tsx
Diogomartf: we can use tailwind top-20 which is 80px, and replace in all of this places.

But I'm okay with having the custom value in this 2 files. It's the height of the nav and here is to show after that nav.

I don't think will use that size any more. So adding to tailwind config is not a good idea.

packages/app/components/Header/MobileMenuList.tsx
```diff
@@ -0,0 +1,38 @@
...
+        onClick={toggle}
+      ></div>
+      <div className="w-full bg-surface-25 rounded-lg absolute top-[72px] left-0 border-b border-solid border-surface-75 gap-2 py-2 px-6">
+        <Link href="#" className="flex py-3 items-center text-med-em">
+          <Image
+            alt="your stacks"
+            src={Blocks}
+            className="w-[18px] h-[18px] mr-2"
```
Diogomartf, line 24: Same comment as above, use the `Image` height and width prop.

packages/app/package.json
Diogomartf: I don't think we need this.

packages/app/components/Header/Header.tsx
```diff
@@ -0,0 +1,47 @@
...
+        </div>
+        <div className="border-r border-solid border-b-gray-100 h-[31px] ml-[20px] md:ml-[40px]"></div>
```
Diogomartf, line 25: **We should avoid using custom sizes! Only use it when is straighly necessary.**

For eg. we have the `h-8` which is `32px` I don't think that `1px` will matter in this case.

By using the tailwind scale we benefit the consistency.

```suggestion
        <div className="border-r border-solid border-b-gray-100 h-8 ml-5 md:ml-10"></div>
```

packages/app/components/Header/MobileMenuList.tsx
berteotti: Nice point, forgot about that. That's why we are using tailwind. Thanks for bringing it up Diogo.

packages/app/components/Header/HamburgerButton.tsx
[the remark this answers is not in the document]
Diogomartf: **Tip**
`p-2` is equal `px-2 py-2`

packages/app/components/Header/HamburgerButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+      onClick={toggle}
+    >
+      {open ? (
+        <Image
+          aria-label="close-menu"
+          alt="close-menu"
+          src={CrossIcon}
+          className="h-[20px] w-[20px] p-[4px]"
```
Diogomartf, line 24: ⚠️ avoid custom sizes ⚠️

Also on the image we can use the `height` and `width` property.

packages/app/components/Header/HamburgerButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+        />
+      ) : (
+        <Image
+          aria-label="menu"
+          alt="menu"
+          src={HamburgerIcon}
+          className="h-[20px] w-[20px]"
```
Diogomartf, line 31: ⚠️ avoid custom sizes ⚠️

On next Image comp we can use the height and width property.

packages/app/components/Header/HamburgerButton.tsx
```diff
@@ -0,0 +1,36 @@
+import HamburgerIcon from "../../public/assets/images/hamburger.svg";
+import CrossIcon from "../../public/assets/images/cross.svg";
```
Diogomartf, line 2: the icon is not a cross, is more an `X` or `close` can we rename it?

packages/app/components/Header/HamburgerButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+import Image from "next/image";
+
+interface HamburgerButtonProps {
+  open: boolean;
+  toggle: () => void;
+}
+
+export default function HamburgerButton({
```
Diogomartf, line 10: Let's rename this component to `MenuButton`, as it not only Hamburguer button. It's the Hamburguer and close icon together.

packages/app/components/Header/Header.tsx
```diff
@@ -0,0 +1,47 @@
+"use client";
+import Image from "next/image";
+import Logo from "../../public/assets/images/logo.svg";
+import Stackly from "../../public/assets/images/stackly.svg";
+import Blocks from "../../public/assets/images/blocks.svg";
+
+import Link from "next/link";
+import MobileMenu from "./MobileMenu";
+
+export default function Header() {
```
Diogomartf, line 10: This should call this component `Navbar`

Diogomartf (changes requested):
Overall good work Velu.

I've made a lot of components but mostly are about `tailwindcss`. 
I think in the upcoming PRs will be better at tailwind, so I took more time on this one.

Let's fix these minor issues and ship the navbar.

packages/app/components/Header/Header.tsx
wixzi: You're importing a component that needs useState. It only works in a Client Component but none of its parents are marked with "use client", so they're Server Components by default.

packages/app/components/Header/HamburgerButton.tsx
wixzi: Sure. :+1:

packages/app/components/Header/HamburgerButton.tsx
wixzi: Ok. Updated

packages/app/package.json
wixzi: https://nextjs.org/docs/messages/install-sharp

packages/app/components/Header/MobileMenuList.tsx
```diff
@@ -0,0 +1,38 @@
...
+          />
+          <span>Your Stacks</span>
+        </Link>
+        <hr className="border-b border-solid border-surface-75 h-0 -ml-6 -mr-6 text-med-em" />
+        <Link href="#" className="block py-3">
+          How it works
+        </Link>
+        <Link href="#" className="block py-3 text-med-em">
```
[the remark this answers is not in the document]
wixzi: I was using the text color here. :) `text-text-med-em` from tailwind.config -> colors-> text

wixzi:
> Overall good work Velu.
> 
> I've made a lot of components but mostly are about `tailwindcss`. I think in the upcoming PRs will be better at tailwind, so I took more time on this one.
> 
> Also adding a small description of the work and screenshot in the PR description would be awesome.
> 
> Let's fix these minor issues and ship the navbar.

Thank you. It's been some time since I used Tailwind. Going forward this issue will not occur.

packages/app/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,35 @@
+"use client";
+import Image from "next/image";
+import Blocks from "../../public/assets/images/blocks.svg";
+import Link from "next/link";
+import MobileMenu from "./MobileMenu";
+import Logo from "./Logo";
+
+export default function Header() {
```
Diogomartf, line 8: Let's change the name of the component to `Navbar`
```suggestion
export default function Navbar() {
```

Diogomartf:
Just that small change and I think is ready ✔️

berteotti:
Hey Velu, I advise you to download this vscode extension if you haven't already. It's pretty useful for this cases https://marketplace.visualstudio.com/items?itemName=bradlc.vscode-tailwindcss

Diogomartf (dismissed):

packages/app/components/Navbar/MenuButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+interface MenuButtonProps {
+  open: boolean;
+  toggle: () => void;
+}
+
+export default function MenuButton({ open, toggle }: MenuButtonProps) {
+  return (
+    <button
+      className="space-y-1.5 md:invisible p-2.5 rounded-[12px]  bg-surface-50 active:bg-surface-75"
```
berteotti, line 13: We should use `rounded-xl` here

packages/app/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,35 @@
+"use client";
+import Image from "next/image";
+import Blocks from "../../public/assets/images/blocks.svg";
+import Link from "next/link";
+import MobileMenu from "./MobileMenu";
+import Logo from "./Logo";
+
+export default function Navbar() {
+  return (
+    <header className="h-[72px] top-0 flex flex-col w-full bg-surface-25 border-b border-solid border-b-surface-75">
```
berteotti, line 10: IMO `72px` should be a constant shared between `Navbar` and `MobileMenuList`

berteotti (changes requested):
Let me know what you think about the constant

packages/app/components/Navbar/Navbar.tsx
wixzi: Moved to tailwind extend config.

berteotti (approved):
Good job Velu!

PR #12, opened 2023-05-16
feat: Create Button, ButtonLink and Icon elements

Diogomartf:
## Summary
- Adds `Button` component
  - icon left and right
  - different sizes 
  - disabled, focus, active state
- Adds `ButtonLink` component (shares the same styles with Button but is a link).
- Adds _/ui_ page which will be our gathering of our components
  - to help us display our components, added two helper components for sections
    - adds `UISection` (big section for eg. Buttons)
    - adds `UISubSection`  (to show a subsection of buttons eg action primary)
- Adds Icon component 
  - supports changing color
  - supports different sizes
- Adds Icons section on /ui page

image

Already in use on Navbar

image

Added Icon component and shows all our icons on /ui

image


packages/app/ui/icon/Icon.tsx
```diff
@@ -0,0 +1,67 @@
...
+export type IconName =
+  | "close"
+  | "plus"
+  | "menu"
+  | "blocks"
+  | "check"
+  | "arrow-left"
+  | "arrow-external"
+  | "warning"
+  | "sawp"
```
berteotti, line 24: Typo

berteotti (changes requested):
Build is failing so can't access the preview.

Diogomartf:
> Build is failing so can't access the preview.

build is fixed now

packages/app/ui/icon/Icon.tsx
```diff
@@ -0,0 +1,67 @@
...
+    <Image
+      className={className}
+      aria-label={name}
+      alt={alt ? alt : name}
+      src={getIcon(name)}
+      width={size}
+      height={size}
+    />
+  );
+};
```
berteotti, line 67: I wouldn't use Image for SVG's. It's not possible to style SVG with the current parent color. There's documentation saying that SVG aren't optimized using `Image`, [like this one](https://nextjs.org/docs/pages/api-reference/components/image#dangerouslyallowsvg). I would follow one of this options so we can have better SVG elements https://stackoverflow.com/a/74136395

Diogomartf: Yeah, you're right, I was already looking into that. Thanks.

packages/app/ui/icon/Icon.tsx
```diff
@@ -0,0 +1,69 @@
...
+  warning: WarningIcon,
+  swap: SwapIcon,
+  "caret-down": CaretDownIcon,
+  "caret-left": CaretLeftIcon,
+  "caret-right": CaretRightIcon,
+};
+
+const defaultIcon = iconMap["blocks"];
+
+const getIcon = (name: IconName) => iconMap[name] || defaultIcon;
```
berteotti, line 46: I think the app should break if there isn't the correct Icon. Or at least render a dummy element. But I think throwing an error is better than a dummy element, gives no room for errors.

Diogomartf: Maybe we can remove this,  it's typesafe so it has linting error and will break on compiling.

Diogomartf: addressed here: 074aec04d6e31c699a13ec484e0df09d978537f4

berteotti (approved):
Good job Diogo, looks good to me!

berteotti (approved):

berteotti (approved):

PR #14, opened 2023-05-25
feat: create ChipButton element

berteotti:
This PR adds the new ChipButton and RadioButton UI elements.

Both components are present in UI page.

image

### DIFF
--- a/packages/app/app/ui/page.tsx
+++ b/packages/app/app/ui/page.tsx
@@ -2,9 +2,19 @@
 
 import { UISection } from "@/app/ui/sections/UISection";
 import { UISubSection } from "@/app/ui/sections/UISubSection";
-import { Button, ButtonLink, Icon, IconName, iconMap } from "@/ui";
+import {
+  Button,
+  ButtonLink,
+  Icon,
+  IconName,
+  iconMap,
+  ChipButton,
+  RadioButton,
+} from "@/ui";
+import { useState } from "react";
 
 export default function Page() {
+  const [activeButton, setActiveButton] = useState("0");
   return (
     <div className="my-10">
       <h1 className="text-3xl font-medium">Stackly UI</h1>
@@ -12,7 +22,7 @@ export default function Page() {
         title="Buttons"
         description="This is a collection of all our current buttons divided by actions."
       >
-        <UISubSection title="primary">
+        <UISubSection title="Primary">
           <Button size="lg" onClick={() => console.log("hey")}>
             Try Stackly now
           </Button>
@@ -28,13 +38,13 @@ export default function Page() {
             active={true}
             onClick={() => console.log("hey")}
           >
-            active
+            Active
           </Button>
           <Button size="sm" disabled={true} onClick={() => console.log("hey")}>
-            button disabled
+            Button disabled
           </Button>
         </UISubSection>
-        <UISubSection title="secondary">
+        <UISubSection title="Secondary">
           <Button
             action="secondary"
             iconLeft="plus"
@@ -50,10 +60,10 @@ export default function Page() {
             onClick={() => console.log("hey")}
           ></Button>
           <ButtonLink action="secondary" size="sm" href="/">
-            soy un ButtonLink
+            Soy un ButtonLink
           </ButtonLink>
         </UISubSection>
-        <UISubSection title="tertiary">
+        <UISubSection title="Tertiary">
           <Button
             action="tertiary"
             iconLeft="plus"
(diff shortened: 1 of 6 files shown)


Diogomartf:
the difference between active and disabled is very subtle. Don't know if would be better to add more constrast? What do you think?

image

packages/app/ui/buttons/RadioButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export const radioButtonStyles = cva(
+  [
+    "flex items-center justify-center",
+    "cursor-pointer select-none font-medium",
+    "active:ring-4",
+    "disabled:bg-surface-75 disabled:text-em-disabled disabled:cursor-not-allowed disabled:ring-0",
+    "focus:outline-none focus:ring-4",
+    "rounded-full px-3 py-2 space-x-1.5 text-xs",
```
Diogomartf, line 16: This is just a suggestion. 

I would prefer to move this string up to the second element, near the `flex items-center`.
I like to have the padding above.

Also `text-xs` could be near `font-medium` as they are relatable as well.

Not important, just sharing a preference.

packages/app/ui/buttons/RadioButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+        value={value}
+        id={id}
+        onChange={onChange}
+        checked={checked}
+        name={name}
+        className="w-0 h-0"
+      />
+      <label
+        htmlFor={id}
+        className={cx(
```
Diogomartf, line 59: what styles are you joining here?

I think there is only `radioButtonStyles` so you can remove `cx`.

packages/app/ui/buttons/ChipButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+  onClick,
+  active,
+  id,
+}: ChipButtonProps) {
+  return (
+    <button
+      id={id}
+      onClick={onClick}
+      disabled={disabled}
+      className={twMerge(
```
Diogomartf, line 33: why use `twMerge` with `cx` ?

Diogomartf:
75db7299d99ebbd2d79f22f117c766d0

wondering if it wouldn't be better if the `active ring` when clicking an inactive radio btn was green (primary color) instead of gray?

packages/app/ui/buttons/ChipButton.tsx
berteotti: There were some conflicts happening, so had to use `twMerge` here. `cx` wasn't handling conflicts, also they recommend it on their docs https://cva.style/docs/installation#handling-style-conflicts

packages/app/ui/buttons/RadioButton.tsx
berteotti: Maybe we should have some guidelines to better define how to organize classes.

berteotti:
> the difference between active and disabled is very subtle. Don't know if would be better to add more constrast? What do you think?
> 
> image

In the design case, I think it's desired to be subtle, almost as a "this isn't an option". It's clearly not disabled as the text and cursor show it's enabled. But we can improve. What do you think it should be?

image

packages/app/ui/buttons/RadioButton.tsx
Diogomartf: Maybe we should. We can also try it loose for a couple more weeks and then decide the guidelines.

packages/app/app/ui/page.tsx
```diff
@@ -28,13 +38,13 @@ export default function Page() {
...
-            active
+            Active
           </Button>
           <Button size="sm" disabled={true} onClick={() => console.log("hey")}>
-            button disabled
+            Button disabled
           </Button>
         </UISubSection>
-        <UISubSection title="secondary">
+        <UISubSection title="Secondary">
```
Diogomartf, line 47: maybe `UISection` and `UISubSection` should have the style capitalize instead of us making it.

berteotti:
> 75db7299d99ebbd2d79f22f117c766d0
> 
> wondering if it wouldn't be better if the `active ring` when clicking an inactive radio btn was green (primary color) instead of gray?

I think the active should reflect the current state before clicking. Mixing colors doesn't sound good to me.

packages/app/app/ui/page.tsx
berteotti: It's just a detail. I like it capitalize but we can also have it normal. I think the components should control it, so we can have the freedom to do it as we want :)

Diogomartf:
> > the difference between active and disabled is very subtle. Don't know if would be better to add more constrast? What do you think?
> > image
> 
> In the design case, I think it's desired to be subtle, almost as a "this isn't an option". It's clearly not disabled as the text and cursor show it's enabled. But we can improve. What do you think it should be?
> 
> image

Don't have a good suggestion right now, let it be, and we see if it works in the wild and if not we improve it in the future.

Diogomartf (approved):
Good work Leo 👍

PR #19, opened 2023-06-06
feat: add Stacktable base using mockdata

Diogomartf:
- add Table to ui (shadcn ui)
- add TextDivider component
- add time utils

### DIFF
--- a/packages/app/app/stacks/components/StacksTable.tsx
+++ b/packages/app/app/stacks/components/StacksTable.tsx
@@ -1,0 +1,181 @@
+import { TextDivider } from "./TextDivider";
+import { Order } from "@/app/stacks/page";
+import {
+  BodyText,
+  Button,
+  CaptionText,
+  Table,
+  TableBody,
+  TableCell,
+  TableHead,
+  TableHeader,
+  TableRow,
+} from "@/ui";
+import { currentTimestampInSeconds } from "@/utils/time";
+
+const mockCowOrders = [
+  {
+    creationDate: "2023-06-01T17:47:21.439282Z",
+    owner: "0x68b57d8b652aee685c6b8c387616911eca5ed883",
+    uid: "0x86c95122ed312f02fa9904e8b9d5b19c70317eac0dc0a207bf4d4b36e6686f2668b57d8b652aee685c6b8c387616911eca5ed8836478e024",
+    availableBalance: null,
+    executedBuyAmount: "182201661599561308717",
+    executedSellAmount: "13783979663205039289",
+    executedSellAmountBeforeFees: "13783979663205039289",
+    executedFeeAmount: "0",
+    invalidated: false,
+    status: "fulfilled",
+    class: "limit",
+    surplusFee: "4855798329137440",
+    surplusFeeTimestamp: "2023-06-01T17:47:25.339953Z",
+    executedSurplusFee: "4855798329137440",
+    settlementContract: "0x9008d19f58aabd9ed0d60971565aa8510560ab41",
+    fullFeeAmount: "4855798329137440",
+    solverFee: "4855798329137440",
+    isLiquidityOrder: false,
+    sellToken: "0xe91d153e0b41518a2ce8dd3d7944fa863463a97d",
+    buyToken: "0x177127622c4a00f3d409b75571e12cb3c8973d3c",
+    receiver: "0x05a4ed2367bd2f0aa63cc14897850be7474bc722",
+    sellAmount: "13783979663205039289",
+    buyAmount: "1",
+    validTo: 1685643300,
+    appData:
+      "0x9b5c6dfa0fa4be89e17700f05bee8775b281aa6d2dac7dfbf3945e0f9642d777",
+    feeAmount: "0",
+    kind: "sell",
+    partiallyFillable: false,
+    sellTokenBalance: "erc20",
+    buyTokenBalance: "erc20",
+    signingScheme: "eip1271",
+    signature:
+      "0x000000000000000000000000e91d153e0b41518a2ce8dd3d7944fa863463a97d000000000000000000000000177127622c4a00f3d409b75571e12cb3c8973d3c00000000000000000000000005a4ed2367bd2f0aa63cc14897850be7474bc722000000000000000000000000000000000000000000000000bf4a8879188b7cb90000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000006478e0249b5c6dfa0fa4be89e17700f05bee8775b281aa6d2dac7dfbf3945e0f9642d7770000000000000000000000000000000000000000000000000000000000000000f3b277728b3fee749481eb3e0b3b48980dbbab78658fc419025cb16eee34677500000000000000000000000000000000000000000000000000000000000000005a28e9363bb942b639270062aa6bb295f434bcdfc42c97267bf003f272060dc95a28e9363bb942b639270062aa6bb295f434bcdfc42c97267bf003f272060dc9",
+    interactions: {
+      pre: [],
+      post: [],
+    },
+  },
+];
+
+// mock function to be replaced
+const getCowOrders = (order: Order) => mockCowOrders;
+
+// we'll probably use viem/ethers function when we connect to data
+const convertedAmount = (amount: string | number, decimals: number) =>
+  Number(amount) / 10 ** decimals;
+
+const calculateAveragePrice = (order: Order) => {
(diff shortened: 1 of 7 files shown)


packages/app/app/stacks/page.tsx
```diff
@@ -3,10 +3,185 @@
...
+  interval: string;
+}
+const completedOrders = (orders: Order[]) =>
+  orders.filter(
+    (order) =>
+      Number(order.endTime) < currentTimestampInSeconds &&
+      order.cancelledAt === null
+  );
+
+const currentOrders = (orders: Order[]) =>
```
berteotti, line 176: `activeOrders` is a better naming IMO

packages/app/ui/table/Table.tsx
```diff
@@ -0,0 +1,117 @@
+import * as React from "react";
```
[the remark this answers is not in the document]
Diogomartf: we can remove it but we do need to add `forwardRef` imports. As it came like this from `shadcn` I let be, but we can change.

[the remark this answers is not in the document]
Diogomartf: fixed: [2b34933](https://github.com/SwaprHQ/stackly-ui/pull/19/commits/2b34933df749e7a780ec43f5ff38401a4b4a6b08)

packages/app/app/stacks/components/TextDivider.tsx
```diff
@@ -0,0 +1,14 @@
...
+  left,
+  right,
+}: {
+  left: string;
+  right: string;
+}) => (
+  <div className="flex items-center justify-end space-x-1 min-w-max">
+    <BodyText className="text-em-high">{left}</BodyText>
+    <BodyText className="text-em-low">{right}</BodyText>
+  </div>
```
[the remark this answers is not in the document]
Diogomartf: Thanks for the feedback. I'll try to improve it.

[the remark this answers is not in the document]
Diogomartf: TableCell has other styles, I would not mess with that. 
I think we can have it without the component for now.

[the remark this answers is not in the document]
Diogomartf: something like this:
```js
import { PropsWithChildren } from "react";

export const TextWrapper = ({ children }: PropsWithChildren) => (
  <div className="flex items-center justify-end space-x-1 min-w-max">
    {children}
  </div>
);

```

packages/app/app/stacks/components/StacksTable.tsx
```diff
@@ -0,0 +1,191 @@
...
+    totalExecutedSellAmount += convertedAmount(
+      cowOrder.executedSellAmount,
+      order.sellToken.decimals
+    );
+  });
+  const averagePrice = totalExecutedSellAmount / totalExecutedBuyAmount;
+  return averagePrice;
+};
+
+export const ordersDone = (order: Order) => {
```
berteotti, line 86:
```suggestion
export const totalOrdersDone = (order: Order) => {
```

Diogomartf: I this as:
`totalOrders` = total number of orders
`ordersDone` = number of orders done

Adding total doesn't make sense in my opinion.

berteotti: I suggested total in the sense total equals a number. `totalOrders` is good because I know it returns a number, but `ordersDone` seems to return the orders that are already done

packages/app/app/stacks/components/StacksTable.tsx
```diff
@@ -0,0 +1,191 @@
...
+              <div className="ml-3 space-y-0.5">
+                <BodyText weight="bold">
+                  {totalStacked(order).toFixed(3)}
+                </BodyText>
+                <CaptionText className="text-em-low">
+                  {getPairSymbols(order)}
+                </CaptionText>
+              </div>
+            </TableCell>
+            <TableCell className="text-right">
```
berteotti, line 146: I thought I had comment but I didn't. Should we had `CellWrapper` styles into `TableCell`? I feel that having an extra component to add style isn't a good tradeoff. Or at least we could create a TableCell component with styles, no need for an extra div IMO. Also, if this component is only used here we can declare it in this file :)

packages/app/app/stacks/components/StacksTable.tsx
Diogomartf: interesting, that makes sense, didn't see it that way. Thanks for explaining deeper.

Now I can see why `totalOrdersDone` make sense.

I'll change it to `totalOrdersDone`

packages/app/app/stacks/components/StacksTable.tsx
Diogomartf: There are some `TableCell`'s that we do not want the same styles for eg:

```js
<TableCell className="flex items-center font-medium">
  <StackedTokenLogoPair order={order} />
    <div className="ml-3 space-y-0.5">
      <BodyText weight="bold">{totalStacked(order).toFixed(3)}</BodyText>
      <CaptionText className="text-em-low">
         {getOrderPairSymbols(order)}
       </CaptionText>
    </div>
  </TableCell>
```

packages/app/app/stacks/components/StacksTable.tsx
Diogomartf: I have organised this on the Stack Modal PR so I'll change there if it is ok

packages/app/app/stacks/components/StacksTable.tsx
Diogomartf: Also adding flex to a table `td` element breaks the table layout, and we do not want that.

image

packages/app/app/stacks/components/StacksTable.tsx
Diogomartf: updated here: https://github.com/SwaprHQ/stackly-ui/pull/23/commits/f738ea744b0405f142fc55c1002e5d0e73db32d6

berteotti:
I have two suggestion:

1. Move to `StacksTable.tsx` every component that we are using. The component can be defined inside or outside the parent component. Why? Because we are not reusing the component elsewhere, make it easier to read and work with the component. Also it's easier to move to a file if it's being used elsewhere then moving it inside a file when we realize it's not being use outside `StacksTable`.
2. Move to `components` folder, outside `app`. Why? We are defining other components there, such `Navbar` and `Stackbox`. This will create confusion in the future. Where should a new component be defined? Inside the page folder? Inside components? We can also create components folder inside each page but seems to me that's a less flexible option. When should we create a component inside component folder? "Always" is an easier answer than "if x, y & z".

packages/app/app/stacks/components/StacksTable.tsx
Diogomartf: Interesting, I would take this, but the problem here is that it doesn't handle the mobile styles properly.

image

Diogomartf:
> I have two suggestion:
> 
> 1. Move to `StacksTable.tsx` every component that we are using. The component can be defined inside or outside the parent component. Why? Because we are not reusing the component elsewhere, make it easier to read and work with the component. Also it's easier to move to a file if it's being used elsewhere then moving it inside a file when we realize it's not being use outside `StacksTable`.
> 2. Move to `components` folder, outside `app`. Why? We are defining other components there, such `Navbar` and `Stackbox`. This will create confusion in the future. Where should a new component be defined? Inside the page folder? Inside components? We can also create components folder inside each page but seems to me that's a less flexible option. When should we create a component inside component folder? "Always" is an easier answer than "if x, y & z".

interesting ideas, I think you got a point. I'll update those in the Stack Modal PR.

Diogomartf:
> I have two suggestion:
> 
> 1. Move to `StacksTable.tsx` every component that we are using. The component can be defined inside or outside the parent component. Why? Because we are not reusing the component elsewhere, make it easier to read and work with the component. Also it's easier to move to a file if it's being used elsewhere then moving it inside a file when we realize it's not being use outside `StacksTable`.
> 2. Move to `components` folder, outside `app`. Why? We are defining other components there, such `Navbar` and `Stackbox`. This will create confusion in the future. Where should a new component be defined? Inside the page folder? Inside components? We can also create components folder inside each page but seems to me that's a less flexible option. When should we create a component inside component folder? "Always" is an easier answer than "if x, y & z".

updated here: [c0e6fff](https://github.com/SwaprHQ/stackly-ui/pull/23/commits/c0e6fff1b855d383ab64d20009eb0b79ee4afda4)

PR #21, opened 2023-06-07
feat: add connect button

berteotti:
This PR adds new connect button.

Screenshot 2023-06-06 at 17 20 10


Diogomartf:
Design looks good on Desktop 👌 
image

But it does not show on mobile.
image

Diogomartf:
Some problems I encountered in quick testing on Brave Browser:
- it does not detect metamask, wasn't able to connect
- wasn't able to connect with taho because I didn't have the gnosis network there. After adding it manually it worked.
  - the error wasn't good, it didn't tell me what was wrong, just that connection failed
  
 Would be great:
 - when you are on a not supported network
   - show switch to network
 - when wallet does not have a supported network
   -  add network option

.gitignore
```diff
@@ -13,4 +13,6 @@ tsconfig.tsbuildinfo
 !.yarn/releases
 !.yarn/plugins
 !.yarn/sdks
-!.yarn/versions
\ No newline at end of file
+!.yarn/versions
+
+.env.local
```
Diogomartf, line 18: you should also add `.env`

packages/app/.env
Diogomartf, line 1: this file shouldn't be commited.

We can have a `.env.example` to share the keys.

packages/app/app/page.tsx
```diff
@@ -1,5 +1,11 @@
 import { Stackbox } from "@/components/stackbox/Stackbox";
 
+export const metadata = {
+  title: "Stackly | Stack crypto over time.",
+  description:
+    "Stackly is a simple, non-custodial tool that uses the CoW protocol to place recurring swaps based on DCA..",
+};
```
Diogomartf, line 7: this should be on route layout to be shared across all the app (overridden when we want).
Being on the app page (aka home page) will not be available on other pages we might have.

packages/app/next.config.js
```diff
@@ -3,6 +3,8 @@ const nextConfig = {
   output: "export",
   distDir: "../../dist",
   webpack(config) {
+    config.resolve.fallback = { "pino-pretty": false, lokijs: false };
```
Diogomartf, line 6: why do we need this config?

packages/app/app/layout.tsx
```diff
@@ -1,19 +1,41 @@
...
+    alchemyId: process.env.ALCHEMY_KEY, // or infuraId
+    walletConnectProjectId: "", //process.env.WALLETCONNECT_PROJECT_ID,
+
+    // Required
+    appName: "Stackly",
+
+    // Optional
+    appDescription: "Stack crypto over time.",
+    appUrl: "https://stackly.eth.limo",
+    //appLogo: "https://stackly.eth.limo/favicon.ico", // your app's logo,no bigger than 1024x1024px (max. 1MB)
```
Diogomartf, line 24: might be interesting to add our `appLogo`.

packages/app/app/layout.tsx
```diff
@@ -1,19 +1,41 @@
...
+);
 
 export default function RootLayout({ children }: PropsWithChildren) {
   return (
     <html lang="en">
       <body className="max-w-screen-xl px-4 mx-auto font-sans bg-surface-25">
-        <Navbar />
-        {children}
+        <WagmiConfig config={config}>
+          <ConnectKitProvider>
```
Diogomartf, line 34: maybe we can have a `Providers.tsx` file and extract this code and use only `<Providers></Providers>`

packages/app/app/layout.tsx
```diff
@@ -1,19 +1,41 @@
...
+    // Required
+    appName: "Stackly",
+
+    // Optional
+    appDescription: "Stack crypto over time.",
+    appUrl: "https://stackly.eth.limo",
+    //appLogo: "https://stackly.eth.limo/favicon.ico", // your app's logo,no bigger than 1024x1024px (max. 1MB)
+    chains,
+  })
+);
```
Diogomartf, line 27: Don't feel this code belongs here.

Maybe we can have a providers file? or something like that and extract this code there.

packages/app/components/ConnectButton.tsx
```diff
@@ -0,0 +1,94 @@
...
+    address: address,
+  });
+
+  const truncatedAddress = (size: number) =>
+    `${address.slice(0, size)}...${address.slice(
+      address.length - size,
+      address.length
+    )}`;
+
+  const formattedBalance = (balanceData: NonNullable<typeof balance>) =>
```
Diogomartf, line 31: doesn't `viem` has formating helpers?

.gitignore
berteotti: Don't know if we want to have some shared variables in .env. That's why I created my own .env.local. What do you think?

packages/app/app/page.tsx
berteotti: We have got some conflicts with "use client". Don't know how to solve this in the best way.

packages/app/next.config.js
berteotti: This libs were throwing some warning in the dev server. Found this solution to prevent fix it.

packages/app/app/layout.tsx
berteotti: Were should we add this "component"? Components folder? Providers folder?

packages/app/components/ConnectButton.tsx
berteotti: Didn't find so custom formatting. Will try to find something, otherwise will stick with this solution for now.

packages/app/next.config.js
```diff
@@ -11,6 +13,9 @@ const nextConfig = {
 
     return config;
   },
+  images: {
+    unoptimized: true,
```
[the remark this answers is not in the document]
berteotti: Had one error using next `Image` for the avatar. So this was to solution to fix it. Will see if I can use img element, but it was throwing another warning.

berteotti:
> Member

Network selector will be added in a new PR

packages/app/app/layout.tsx
berteotti: Moved to providers folder

packages/app/components/ConnectButton.tsx
berteotti: There isn't anything to format the balance like we need.

packages/app/app/layout.tsx
berteotti: Updated, but didn't see any changes.

packages/app/.env
Diogomartf, line 1: You have to remove this from git.

I think one way you can do it is to rename the file to `.temp` do `git add`, and then change it again to `.env `

packages/app/components/ConnectButton.tsx
```diff
@@ -0,0 +1,94 @@
...
+        <BodyText size={2} className="ml-3 text-em-med hidden md:block">
+          {formattedBalance(balance)}
+        </BodyText>
+      )}
+      <Button
+        action="tertiary"
+        iconRight="caret-down"
+        onClick={onClick}
+        size="sm"
+        className="border-none shadow-sm rounded-xl flex flex-row hover:bg-white focus:bg-white focus:ring-0 active:ring-0"
```
Diogomartf, line 51: flex row is the default, no need for explicit class.

```suggestion
        className="border-none shadow-sm rounded-xl flex  hover:bg-white focus:bg-white focus:ring-0 active:ring-0"
```

packages/app/components/ConnectButton.tsx
```diff
@@ -0,0 +1,94 @@
...
+        {avatar && (
+          <Image
+            width={20}
+            height={20}
+            src={avatar}
+            alt={address}
+            className="rounded-full"
+          />
+        )}
+        <BodyText size={2} className=" text-black">
```
Diogomartf, line 62: Extra spacing. Not mandatory to remove but I noticed so I'll comment.

```suggestion
        <BodyText size={2} className="text-black">
```

packages/app/next.config.js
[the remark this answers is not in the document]
Diogomartf: Strange ..

packages/app/providers/wagmi-config.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const chains = [gnosis, mainnet];
+
+export const config = createConfig(
+  getDefaultConfig({
+    // Required API Keys
+    alchemyId: process.env.ALCHEMY_KEY, // or infuraId
+    walletConnectProjectId: "", //process.env.WALLETCONNECT_PROJECT_ID,
+
+    // Required
```
Diogomartf, line 13: I would remove this comment

packages/app/providers/wagmi-config.ts
```diff
@@ -0,0 +1,22 @@
...
+import { getDefaultConfig } from "connectkit";
+import { gnosis, mainnet } from "wagmi/chains";
+
+export const chains = [gnosis, mainnet];
+
+export const config = createConfig(
+  getDefaultConfig({
+    // Required API Keys
+    alchemyId: process.env.ALCHEMY_KEY, // or infuraId
+    walletConnectProjectId: "", //process.env.WALLETCONNECT_PROJECT_ID,
```
Diogomartf, line 11: I would remove this comment

packages/app/providers/wagmi-config.ts
```diff
@@ -0,0 +1,22 @@
+import { createConfig } from "wagmi";
+import { getDefaultConfig } from "connectkit";
+import { gnosis, mainnet } from "wagmi/chains";
+
+export const chains = [gnosis, mainnet];
+
+export const config = createConfig(
+  getDefaultConfig({
+    // Required API Keys
```
Diogomartf, line 9: I would remove this comment

packages/app/app/ui/page.tsx
```diff
@@ -123,6 +124,11 @@ export default function Page() {
           </Button>
         </UISubSection>
       </UISection>
+      <UISection title="Connect Button">
+        <div className="flex flex-wrap space-x-4">
+          <ConnectButton />
```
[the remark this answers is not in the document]
berteotti: removed it

packages/app/components/ConnectButton.tsx
berteotti: fixed

packages/app/components/ConnectButton.tsx
berteotti: fixed

packages/app/app/ui/page.tsx
```diff
@@ -123,6 +124,11 @@ export default function Page() {
           </Button>
         </UISubSection>
       </UISection>
+      <UISection title="Connect Button">
+        <div className="flex flex-wrap space-x-4">
```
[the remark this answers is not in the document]
berteotti: removed it

packages/app/components/ConnectButton.tsx
```diff
@@ -0,0 +1,94 @@
+"use client";
+
+import { ConnectKitButton } from "connectkit";
+import { BodyText, Button } from "../ui";
```
Diogomartf: Why not use @/ui like in other files?

I think it's cleaner and it's not dependent where the file is.

Diogomartf (approved):

PR #22, opened 2023-06-09
feat: add select network

berteotti:
Screenshot 2023-06-09 at 11 37 39

We are using select from @headless/ui


Diogomartf:
We do already have a library (@headlessui) that handles selects. 

https://headlessui.com/react/menu

Not sure if we want to add radix just for this use case.

berteotti:
> We do already have a library (@headlessui) that handles selects.
> 
> https://headlessui.com/react/menu
> 
> Not sure if we want to add radix just for this use case.

You are right. Will give @headlessui a try

Diogomartf:
Looks good! 

It would be great to show on hover a slightly highlighted background and cursor pointer.

packages/app/app/layout.tsx
```diff
@@ -1,19 +1,16 @@
...
+import { Navbar } from "@/components";
+import { Providers } from "@/providers";
 import "../styles/global.css";
-import Navbar from "@/components/navbar/Navbar";
-
-export const metadata = {
-  title: "Stackly | Stack crypto over time.",
-  description:
-    "Stackly is a simple, non-custodial tool that uses the CoW protocol to place recurring swaps based on DCA..",
-};
```
Diogomartf, line 9: I think this is important to have on this layout file. This gives a default title and description to all dapp. Overridden on the files we want.

packages/app/components/ConnectButton.tsx
```diff
@@ -0,0 +1,102 @@
...
+        <BodyText size={2} className="ml-3 text-em-med hidden md:block">
+          {formattedBalance(balance)}
+        </BodyText>
+      )}
+      <Button
+        action="tertiary"
+        iconRight="caret-down"
+        onClick={onClick}
+        size="sm"
+<<<<<<< HEAD
```
Diogomartf, line 51: GIT  !!! 🚨

packages/app/components/ConnectButton.tsx
```diff
@@ -0,0 +1,102 @@
...
+        className="border-none shadow-sm rounded-xl flex hover:bg-white focus:bg-white focus:ring-0 active:ring-0"
+=======
+        className="border-none shadow-sm rounded-xl flex flex-row hover:bg-white focus:bg-white focus:ring-0 active:ring-0"
+>>>>>>> 274259e (feat: add connect button)
```
Diogomartf, line 55: GIT  !!! 🚨

Diogomartf:
changing network on my the selector doesn't change on taho wallet.

Diogomartf:
nice to have on select and connect button.

A subtle background change when hovering

image

berteotti:
> 

I don't like connect button hover effect, the balance section mixes with the button, creating what seems a bigger button, but isn't. 

Screenshot 2023-06-16 at 11 35 40

I wouldn't add it to the connect button, but network button can have such effect. Should we do it on network only, both, or none?

Diogomartf:
> > 
> 
> I don't like connect button hover effect, the balance section mixes with the button, creating what seems a bigger button, but isn't.
> 
> Screenshot 2023-06-16 at 11 35 40
> 
> I wouldn't add it to the connect button, but network button can have such effect. Should we do it on network only, both, or none?

If they don't use the same base, we can add it only to the network selector for now.
If they do, we can skip adding the hover effect.

berteotti:
> > > 
> > 
> > 
> > I don't like connect button hover effect, the balance section mixes with the button, creating what seems a bigger button, but isn't.
> > Screenshot 2023-06-16 at 11 35 40
> > I wouldn't add it to the connect button, but network button can have such effect. Should we do it on network only, both, or none?
> 
> If they don't use the same base, we can add it only to the network selector for now. If they do, we can skip adding the hover effect.

They can but they don't, I think we need to be flexible in the beginning

packages/app/components/SelectNetwork.tsx
```diff
@@ -0,0 +1,73 @@
...
+          as={Fragment}
+          leave="transition ease-in duration-100"
+          leaveFrom="opacity-100"
+          leaveTo="opacity-0"
+        >
+          <Listbox.Options className="absolute mt-1 max-h-60 w-auto overflow-auto rounded-2xl bg-white py-1 text-base shadow-md  focus:outline-none sm:text-sm">
+            {chains.map(({ id, name }) => (
+              <Listbox.Option
+                key={id}
+                className="relative cursor-pointer select-none py-2 pr-10 pl-4 hover:bg-surface-75"
```
Diogomartf, line 45: This hover is too strong, please use the surface-25 for less contrast.

```suggestion
                className="relative cursor-pointer select-none py-2 pr-10 pl-4 hover:bg-surface-25"
```

Diogomartf: I was wrong, cause the surface-25 is the bg, maybe we should keep it 75 for now. Sorry!

Diogomartf (approved):

PR #23, opened 2023-06-09
feat: add stack order modal

Diogomartf:
## Description
This PR adds stack modal to stacks table.

### summary
- add stack modal
  - order details 
  - orders progress bar
- add models folder to handle model related functions and types
  -  add order model 
- add utils folder
  - time utils
  - token utils
- updates Modal component
  - sharing type props
  - better spacing
  - handles width better with max-width
  
  
**Screenshot**
  
image


packages/app/app/models/order/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./order";
```
berteotti, line 1: Should we move this folder `models` outside the `app`? This looks something that can live in the same level as components

packages/app/components/OrdersProgressBar.tsx
```diff
@@ -0,0 +1,20 @@
+import { OrderProps, totalOrdersDone, totalOrders } from "@/app/models/order";
+import React, { useRef, useEffect } from "react";
+
+export const OrdersProgressBar = ({ order }: OrderProps) => {
```
berteotti, line 4: Should we do the same with this component? Seems to be a very custom component that can be defined inside `StackProgress`.

packages/app/app/models/order/index.ts
berteotti: Also this is shared between pages and components. So makes senso to live in the same level.

packages/app/components/OrdersProgressBar.tsx
```diff
@@ -0,0 +1,20 @@
...
+  const progressBarRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (progressBarRef.current) {
+      const totalWidth = progressBarRef.current.offsetWidth;
+      const width = (totalWidth * totalOrdersDone(order)) / totalOrders(order);
+      progressBarRef.current.style.width = `${width}px`;
+    }
+  }, [order]);
```
berteotti, line 13: Isn't calculating a percentage with better than this? No need for refs, and we can use tailwind to define the custom percentage width.

packages/app/components/stack-modal/StackProgress.tsx
```diff
@@ -0,0 +1,44 @@
...
+    <div className="flex flex-col justify-between space-y-1 md:space-y-0 md:items-center md:flex-row">
+      <OrdersExecuted order={order} />
+      <div className="flex items-center space-x-1">
+        <BodyText size="responsive" className="text-em-low">
+          Total funds used:{" "}
+          <span className="text-em-high">
+            {totalFundsUsed(order).toFixed(2)}{" "}
+            <span className="text-xs text-em-low">of</span>{" "}
+            {fundsUsedWithToken(order)}
+          </span>
```
berteotti, line 21: This looks a bit different from the design, any reason in particular?

packages/app/components/stack-modal/StackProgress.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { OrdersProgressBar } from "@/components/OrdersProgressBar";
+import { BodyText } from "@/ui";
+
+export const StackProgress = ({ order }: OrderProps) => (
+  <div className="mt-3 space-y-3">
+    <div className="flex flex-col justify-between space-y-1 md:space-y-0 md:items-center md:flex-row">
+      <OrdersExecuted order={order} />
+      <div className="flex items-center space-x-1">
+        <BodyText size="responsive" className="text-em-low">
+          Total funds used:{" "}
```
berteotti, line 16:
```suggestion
          Total Funds used:{" "}
```

I like how the design looks :)

packages/app/components/stack-modal/StackProgress.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <div className="w-4 h-4 rounded-full bg-primary-100"></div>
+      </div>
+    </div>
+    <OrdersProgressBar order={order} />
+  </div>
+);
+
+const OrdersExecuted = ({ order }: OrderProps) => {
+  if (!totalOrdersDone(order))
+    return <BodyText className="text-em-low">No orders executed yet</BodyText>;
```
berteotti, line 32:
```suggestion
    return <BodyText className="text-em-low">No orders executed yet.</BodyText>;
```

Just a subtle change.

packages/app/utils/time.ts
```diff
@@ -1,8 +1,13 @@
-export const currentTimestampInSeconds = Math.floor(Date.now() / 1000);
```
berteotti, line 1: Suggestion for this file name: `datetime.ts` or `date-time.js`

berteotti (changes requested):
Looks good, just left some comments :)

packages/app/components/stack-modal/StackProgress.tsx
Diogomartf: Fixed. thks

packages/app/utils/time.ts
Diogomartf: 👍

Diogomartf: datetime.js (I think in this case makes sense to be together)

packages/app/components/stack-modal/StackProgress.tsx
Diogomartf: was that the difference you noticed?

packages/app/components/OrdersProgressBar.tsx
Diogomartf: Actually this one will be used on other parts, that's why I add it to components. For eg. when we do the custom OrderCards on mobile.

image

packages/app/components/stack-modal/StackProgress.tsx
Diogomartf: I don't think that capitalizing funds make sense. I prefer: Total funds used:

packages/app/app/models/order/index.ts
Diogomartf: fixed: https://github.com/SwaprHQ/stackly-ui/pull/23/commits/d950260c05cfac42e9a6372d7572331783ca866b

packages/app/components/OrdersProgressBar.tsx
Diogomartf: I tried at runtime but couldn't make it work with tailwind custom properties. Maybe I'm missing something. 

But from a quick search:

> Tailwind doesn’t include any sort of client-side runtime, so class names need to be statically extractable at build-time, and can’t depend on any sort of arbitrary dynamic values that change on the client

packages/app/components/stack-modal/StackProgress.tsx
berteotti: I think it's a reading hack, where you put more emphasis on what matters (which is Total Funds), but it works in both ways for me.

packages/app/components/stack-modal/StackProgress.tsx
Diogomartf: yeah, this part is getting big and the "of" is not important, so I made it smaller and decreased the emphasis.

image

packages/app/components/stack-modal/StackProgress.tsx
Diogomartf: I bite this. Updated.

berteotti (approved):

PR #24, opened 2023-06-19
Stk 29 create tokenicon component

Diogomartf:
# Description
This PR adds `TokenIcon` component which fetches token image from a token list based on address.
If it does not have it shows a default component with token symbol.

image

### Summary
Add `TokenIcon` component
  - Gets token list based on chain
    - uses honeyswap list on gnosis
    - uses  1inch list on ethereum
    - Tested out with `coingecko` but assets are small and blury.
  - get image url for a given token address from the list
  - if does not have data for that token, shows a default component
- Updates types/react and typescript packages.

### DIFF
--- a/packages/app/components/FromToStackTokenPair.tsx
+++ b/packages/app/components/FromToStackTokenPair.tsx
@@ -1,6 +1,7 @@
 import { Order, Token, totalFundsUsed } from "@/models/order";
 import { totalStacked } from "@/components/StacksTable";
 import { Icon, TitleText } from "@/ui";
+import { TokenIcon } from "@/components/TokenIcon";
 
 interface FromToStackTokenPairProps {
   order?: Order;
@@ -12,22 +13,27 @@ export const FromToStackTokenPair = ({
   fromToken,
   toToken,
   order,
-}: FromToStackTokenPairProps) => (
-  <div className="flex items-center space-x-4 rounded-3xl w-fit">
-    <div className="flex items-center space-x-2">
-      <div className="w-6 h-6 rounded-full bg-primary-100"></div>
-      <TitleText>
-        {order ? totalFundsUsed(order).toFixed(2) : fromToken?.symbol}
-      </TitleText>
+}: FromToStackTokenPairProps) => {
+  const token0 = order ? order?.sellToken : fromToken;
+  const token1 = order ? order?.buyToken : toToken;
+
+  return (
+    <div className="flex items-center space-x-4 rounded-3xl w-fit">
+      <div className="flex items-center space-x-2">
+        {token0 && <TokenIcon token={token0} size="sm" />}
+        <TitleText>
+          {order ? totalFundsUsed(order).toFixed(2) : fromToken?.symbol}
+        </TitleText>
+      </div>
+      <div className="p-1 rounded-xl md:rounded-2xl md:p-2 bg-surface-75">
+        <Icon size={20} className="rotate-180" name="arrow-left" />
+      </div>
+      <div className="flex items-center space-x-2">
+        {token1 && <TokenIcon token={token1} size="sm" />}
+        <TitleText>
+          {order ? totalStacked(order).toFixed(4) : toToken?.symbol}
+        </TitleText>
+      </div>
     </div>
-    <div className="p-1 rounded-xl md:rounded-2xl md:p-2 bg-surface-75">
-      <Icon size={20} className="rotate-180" name="arrow-left" />
-    </div>
-    <div className="flex items-center space-x-2">
-      <div className="w-6 h-6 rounded-full bg-primary-100"></div>
-      <TitleText>
-        {order ? totalStacked(order).toFixed(4) : toToken?.symbol}
-      </TitleText>
-    </div>
-  </div>
-);
+  );
+};
--- a/packages/app/components/TokenIcon.tsx
+++ b/packages/app/components/TokenIcon.tsx
@@ -1,0 +1,107 @@
+import { isAddress } from "viem";
+import { cva } from "class-variance-authority";
+import { Token } from "@/models/order";
+import Image from "next/image";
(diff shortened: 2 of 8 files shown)


packages/app/components/TokenIcon.tsx
```diff
@@ -0,0 +1,107 @@
...
+export const TokenIcon = ({ token, className, size }: TokenIconProps) => {
+  const { chain } = useNetwork();
+  const [tokensList, setTokensList] = useState<TokenFromList[]>([]);
+
+  const setupTokenList = useCallback(async () => {
+    async function getTokensListData() {
+      const res = await fetch(
+        chain
+          ? TOKEN_LIST_BY_CHAIN_URL[chain.id]
+          : TOKEN_LIST_BY_CHAIN_URL[GNOSIS_CHAIN_ID]
```
berteotti, line 40: Should we fetch if there is no chain?

Diogomartf: What do you mean by no chain? The user isn't connect to a chain?

If so, yes, I think we need to have a default chain when user is not connected.

packages/app/components/FromToStackTokenPair.tsx
```diff
@@ -12,22 +14,27 @@ export const FromToStackTokenPair = ({
   fromToken,
   toToken,
   order,
```
berteotti, line 16: It seems to me that this component is receiving too much stuff. It should receive `to` and `from` token only. The logic should live in the parent component. If we have another source of tokens, should we add it here? If we want to have a component that receives an order we must create `OrderFromToStackTokenPair`.

packages/app/components/TokenIcon.tsx
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export const TokenIcon = ({ token, className, size }: TokenIconProps) => {
+  const { tokenList } = useTokenListContext();
+
+  function getTokenFromList(tokenAddress: string) {
+    return tokenList.find(
+      (el) => el.address.toUpperCase() === tokenAddress.toUpperCase()
+    );
+  }
```
berteotti, line 22: Should this be inside the context?

packages/app/context/TokenListContext.tsx
```diff
@@ -0,0 +1,98 @@
...
+    async function gettokenListData() {
+      const res = await fetch(
+        chain
+          ? TOKEN_LIST_BY_CHAIN_URL[chain.id]
+          : TOKEN_LIST_BY_CHAIN_URL[GNOSIS_CHAIN_ID]
+      );
+      // todo: should we throw an error here or a warning? since it should work without it.
+      if (!res.ok) {
+        throw new Error("Failed to fetch tokenlist data");
+      }
```
berteotti, line 75: We can throw an error but we need to catch it and use it somewhere, like logging it.

packages/app/public/assets/blockchains/ethereum/tokenlist.json
```diff
@@ -0,0 +1,91 @@
+
+[
+  {
+    "address": "0x6cAcDB97e3fC8136805a9E7c342d866ab77D0957",
+    "name": "Swapr on Ethereum mainnet",
+    "symbol": "SWPR",
+    "decimals": 18,
+    "logoURI":
+      "/assets/images/tokens/swapr.png"
```
berteotti, line 9: Why should we have the images locally? Can't we use some link URI?

Diogomartf: Those images were hand curated. No list has good images for Swapr, XDAI, COW and some other tokens.
Having our own images to, what we consider, top tokens is the best way to have control over the design.

packages/app/context/TokenListContext.tsx
Diogomartf: Ok, let's do try catch and log it.

packages/app/components/FromToStackTokenPair.tsx
Diogomartf: Makes sense, will try to simplify this component.

packages/app/context/TokenListContext.tsx
Diogomartf: fixed

packages/app/components/TokenIcon.tsx
Diogomartf: Yeah probably is cleaner. I moved the other function as well.

packages/app/components/FromToStackTokenPair.tsx
Diogomartf: fixed.

packages/app/context/TokenListContext.tsx
```diff
@@ -0,0 +1,111 @@
...
+        )
+    )
+  );
+
+export const TokenListProvider = ({ children }: PropsWithChildren) => {
+  const { chain } = useNetwork();
+
+  const defaultTokenList = chain
+    ? DEFAULT_TOKEN_LIST_BY_CHAIN[chain.id]
+    : DEFAULT_TOKEN_LIST_BY_CHAIN[GNOSIS_CHAIN_ID];
```
berteotti, line 59: I think defaulting to Gnosis chain will bring unexpected behaviour. But we can't test it write now.

berteotti: 1 solution would be to return an empty state if there is no chain.

berteotti (approved):
Looks good to me, just left a comment with 1 concern.

packages/app/context/TokenListContext.tsx
Diogomartf: but we do want a default chain, no?

For eg:

I want to try out Stackly without connecting, which network should be the default for showing tokens?  (token picker), etc.

I think that network, for us, will be gnosis.

PR #29, opened 2023-07-10
[STK-88][FEAT] - TokenPicker SearchBar functionality

ElRodrigote:
### Description

* General cleanup
* Filter out tokens by connected wallet chain
* Implement token search for TokenPicker
  * Full and partial matches
  * Debounced search
  * Empty state + Search Cleanup

image
image
image

### How to test the changes

1) Pull this branch
2) Run the project locally
3) Go to Stackly landing page, click any "select token" button and wait for the modal to pop-up
4) Interact with the Search input, there will be a delay of 0.5s to check results.

### DIFF
--- a/packages/app/components/token-picker/TokenPicker.tsx
+++ b/packages/app/components/token-picker/TokenPicker.tsx
@@ -34,7 +34,7 @@ interface TokenListRowProps {
 
 interface TokenListProps {
   closeAction: () => void;
-  onClearSearchClick: () => void;
+  onClearSearch: () => void;
   tokenList: Token[];
   tokenSearchQuery?: string;
 }
@@ -49,7 +49,7 @@ export const TokenPicker = ({
 
   const { tokenList } = useTokenListContext();
 
-  const tokenListClearSearchCleanup = () => {
+  const tokenListSearchCleanup = () => {
     setDebouncedQuery("");
     setTokenSearchQuery("");
   };
@@ -59,7 +59,7 @@ export const TokenPicker = ({
   };
 
   const handleModalClose = () => {
-    if (debouncedQuery) tokenListClearSearchCleanup();
+    if (debouncedQuery) tokenListSearchCleanup();
     closeAction();
   };
 
@@ -92,7 +92,7 @@ export const TokenPicker = ({
         <CommonTokens />
         <TokenList
           closeAction={handleModalClose}
-          onClearSearchClick={tokenListClearSearchCleanup}
+          onClearSearch={tokenListSearchCleanup}
           tokenList={tokenList}
           tokenSearchQuery={tokenSearchQuery}
         />
@@ -162,7 +162,7 @@ const TokenListRow = ({ closeAction, token }: TokenListRowProps) => (
 
 const TokenList = ({
   closeAction,
-  onClearSearchClick,
+  onClearSearch,
   tokenList,
   tokenSearchQuery
 }: TokenListProps) => {
@@ -170,8 +170,8 @@ const TokenList = ({
     tokenList
   );
 
-  const handleClearSearchClick = () => {
-    onClearSearchClick();
+  const handleClearSearch = () => {
+    onClearSearch();
     setFilteredTokenList(tokenList);
   };
 
@@ -203,11 +203,7 @@ const TokenList = ({
             <div className="flex items-center justify-center flex-col mt-8 space-y-4">
               <EmptyStateImg />
               <BodyText>{`Nothing found for "${tokenSearchQuery}"`}</BodyText>
-              <Button
-                action="secondary"
-                onClick={handleClearSearchClick}
-                size="md"
-              >
+              <Button action="secondary" onClick={handleClearSearch} size="md">
                 Clear search
               </Button>
             </div>


packages/app/public/assets/images/empty-state-stacks-coins.svg
Diogomartf, line 1: They are suppose to be different images.

We're using the first one (deleted) on empty stacks. 

I think the new one is supposed to be more simple for the token picker.

packages/app/components/token-picker/TokenPicker.tsx
```diff
@@ -92,17 +135,58 @@ const TokenListRow = ({ closeAction, token }: TokenListRowProps) => (
...
+    onClearSearchClick();
+    setFilteredTokenList(tokenList);
+  };
+
+  useEffect(() => {
+    if (tokenSearchQuery) {
+      const filteredItems = tokenList.filter(
+        token =>
+          token.symbol.toLowerCase().includes(tokenSearchQuery.toLowerCase()) ||
+          token.name.toLowerCase().includes(tokenSearchQuery.toLowerCase())
```
Diogomartf, line 158: I'm okay with this code but will leave an idea if you find it interesting:


```js
const checkMatch = (matchString, searchQuery) => {
 matchString.toLowerCase().includes(searchQuery.toLowerCase())
}
```

```
checkMatch(token.symbol) || checkMatch(token.name) 
```

something like this would remove some repetition.

Diogomartf:
This PR is looking really good!

Our Modal component can receive a `initialFocusRef` that we can use to focus another element (default is the first clickable one which is the close X), it would be great if could pass so the focus element is the search input.

You can see an example on `ui/page.tsx` with Dialog (which is based on Modal component)

```js
 const dialogBtnRef = useRef<HTMLButtonElement>(null);
 
 <Dialog
    initialFocusRef={dialogBtnRef}
    isOpen={isOpenCancelStackingDialog}
    closeAction={() => setOpenCancelStackingDialog(false)}
  >
  ...
</Dialog>
 ```

**Comment by Diogomartf** [comment]
Small detail 💅 

Also noticed that the font size on the SearchInput is too big, it should be `14px`.

File: packages/app/components/token-picker/TokenPicker.tsx (hunk 1 above)
Comment (berteotti) [reply packages/app/components/token-picker/TokenPicker.tsx:158]: Fine with both. But I would think in a better naming for the function like `matches` or `stringIncludes`.  Also, no function no naming troubles x)

File: packages/app/components/token-picker/TokenPicker.tsx [hunk 2]
```diff
@@ -92,17 +135,58 @@ const TokenListRow = ({ closeAction, token }: TokenListRowProps) => (
...
+            <TokenListRow
+              closeAction={closeAction}
+              key={token.address}
+              token={token}
+            />
+          ))
+        : tokenSearchQuery && (
+            <div className="flex items-center justify-center flex-col mt-8 space-y-4">
+              <EmptyStateImg />
+              <BodyText>{`Nothing found for "${tokenSearchQuery}"`}</BodyText>
```
berteotti, line 180: Love this detail!

packages/app/public/assets/images/empty-state-stacks-coins.svg
ElRodrigote: Oh sorry, my bad, will re-create the old one, the first time I did a lookup for imports on that SVG got no results, maybe I had a typo or something like that. Thanks for the heads up!

packages/app/components/token-picker/TokenPicker.tsx
Diogomartf: yeah, the names can improve! and yeah, agree, bad naming might be better no function!

ElRodrigote: If you don't mind it, I rather leave it as it is right now and abstract it in the future if needed. 

Although I'm a big fan of abstracting repeated code, in this particular case we will end up with more lines of code (because `tokenSearchQuery` may be falsy, thus typechecking workarounds)

ElRodrigote:
> ```js
> const dialogBtnRef = useRef<HTMLButtonElement>(null);
> ```

Ahhh I really like this idea, thanks for the heads-up! Will definitely implement it.

Diogomartf:
I've just noticed that when we search for a token and then close the token picker and open again it the search is not cleared.

I think we want to clear the search every time we close the token picker, what do you think?

Also not sure if it is out of scope of this PR.

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -15,6 +15,8 @@ export const Stackbox = () => {
   const openTokenPicker = () => setTokenPickerIsOpen(true);
   const closeTokenPicker = () => setTokenPickerIsOpen(false);
 
+  const searchBarRef = useRef<HTMLInputElement>(null);
```
Diogomartf, line 18: maybe `searchTokenBar` it's more descriptive?

Diogomartf, line 18: maybe `searchTokenBar` it's more descriptive?

Diogomartf:
> I've just noticed that when we search for a token and then close the token picker and open again it the search is not cleared.
> 
> I think we want to clear the search every time we close the token picker, what do you think?
> 
> Also not sure if it is out of scope of this PR.

What's your thoughts here?

berteotti (approved):
Code looks good to me. Good job Rorry!

packages/app/components/token-picker/TokenPicker.tsx
```diff
@@ -4,51 +4,119 @@ import { useTokenListContext } from "@/context/TokenListContext";
...
+  closeAction,
+  initialFocusRef,
+  isOpen
+}: TokenPickerProps) => {
+  const [tokenSearchQuery, setTokenSearchQuery] = useState("");
+  const [debouncedQuery, setDebouncedQuery] = useState(tokenSearchQuery);
+
   const { tokenList } = useTokenListContext();
 
+  const tokenListClearSearchCleanup = () => {
```
Diogomartf, line 52: I think If we remove the `clear` in the middle it be more clear. (pun intended).

I think CleanUp is already describing that, clear things and stuff.  It feels duplicate.

```suggestion
  const tokenListSearchCleanup = () => {
```

packages/app/components/token-picker/TokenPicker.tsx
```diff
@@ -92,17 +160,58 @@ const TokenListRow = ({ closeAction, token }: TokenListRowProps) => (
...
+  closeAction,
+  onClearSearchClick,
+  tokenList,
+  tokenSearchQuery
+}: TokenListProps) => {
+  const [filteredTokenList, setFilteredTokenList] = useState<Token[]>(
+    tokenList
+  );
+
+  const handleClearSearchClick = () => {
```
Diogomartf, line 173: I would remove the `click` on the name, I don't think it's relevant.

I can see it's a click on the button bellow if I want to. But It's not important to reason about this code.

```suggestion
  const handleClearSearch = () => {
```

packages/app/components/token-picker/TokenPicker.tsx
```diff
@@ -92,17 +160,58 @@ const TokenListRow = ({ closeAction, token }: TokenListRowProps) => (
...
-        closeAction={closeAction}
-        key={token.address}
-        token={token}
-      />
-    ))}
-  </div>
-);
+const TokenList = ({
+  closeAction,
+  onClearSearchClick,
```
Diogomartf, line 165: [On the same note ](https://github.com/SwaprHQ/stackly-ui/pull/29/files#r1262341130)

I would remove click from the name.

```suggestion
  onClearSearch,
```

ElRodrigote:
> > I've just noticed that when we search for a token and then close the token picker and open again it the search is not cleared.
> > I think we want to clear the search every time we close the token picker, what do you think?
> > Also not sure if it is out of scope of this PR.
> 
> What are your thoughts here?

Sorry for leaving this not answered here, but I ended up applying this change :D

Diogomartf (approved):
Great job 💪

PR #30, opened 2023-07-12
feat: connect with subgraph via sdk

Diogomartf:


packages/app/models/order/fetcher.ts
```diff
@@ -1,143 +1,16 @@
...
+import { getUserOrders, getSubgraphEndpoint, ChainId } from "@stackly/sdk";
 
-export async function getOrders(address: string) {
-  return mockOrders.orders;
+export async function getOrders(chainId: ChainId, address: string) {
+  try {
+    const graphqlClient = new GraphQLClient(getSubgraphEndpoint(chainId));
+    const orders = await getUserOrders(graphqlClient, address);
+
+    if (!orders) throw new Error("Failed to fetch SDK data");
```
berteotti, line 10: `"Failed to fetch subgraph data"` is the correct error message.

packages/app/models/order/fetcher.ts
```diff
@@ -1,143 +1,16 @@
...
+
+    return filterAllButCancelledOrders(orders);
```
berteotti, line 12: Can we rename this to `filterCancelledOrders` ?

berteotti: Also, why are we filtering cancelled orders? Shouldn't this orders go to complete orders section at least?

Diogomartf: If we change the name maybe `filterOutCancelledOrders`, because it does not filter cancelled orders, filters all but cancelled.

Diogomartf: But you're right, I think they should go to `complete` tab (at least for now).

packages/app/models/order/fetcher.ts
Diogomartf: If we change the underlying mechanism of the `SDK` this would also work.

So I think this error message is okay, we don't need to be super specific in this case. 

Let me know if you have a strong opinion on this and I'll change it. Otherwise let's keep it.

berteotti: I think we can be more specific. There are 2 ways we can fetch data. With subgraphs or querying contracts. I think we should be specific in this case, it can help debugging in the future. Also, yesterday I learned that we don't need to throw an Error object. We can simply do this:

```suggestion
    if (!orders) throw "Failed to fetch SDK data";
```

Diogomartf: Nice.

Okay, I'll update it then.

packages/app/models/order/types.ts
```diff
@@ -1,20 +1,6 @@
...
-  createdAt: string;
-  startTime: number;
-  endTime: number;
-  orderSlots: string[];
-  cancelledAt: string | null;
-  interval: string;
-}
+import { Order as StacklyOrder } from "@stackly/sdk/dist/vaults/subgraph";
 
+export interface Order extends StacklyOrder {}
```
berteotti, line 3: This is cleaner.

```suggestion
import { Order } from "@stackly/sdk/dist/vaults/subgraph";
```

berteotti: Also, we should not be importing this from `dist` folder. Should come directly from `@stackly/sdk` like ChainId.

packages/app/models/order/types.ts
```diff
@@ -1,20 +1,6 @@
...
 export interface OrderProps {
   order: Order;
 }
```
berteotti, line 6: Should we have a file just for a components prop interface?

berteotti: We are using it more than once so it makes sense.

packages/app/models/order/types.ts
Diogomartf: I don't think it's exported from `SDK`

Diogomartf: Exported type Order from Sdk

packages/app/models/order/types.ts
```diff
@@ -1,4 +1,4 @@
-import { Order as StacklyOrder } from "@stackly/sdk/dist/vaults/subgraph";
+import { Order as StacklyOrder } from "@stackly/sdk";
 
 export interface Order extends StacklyOrder {}
```
berteotti, line 3: Still can't understand why we are doing this. We are importing from `Order` from sdk, renaming it to `StacklyOrder`, extending `StacklyOrder` to a new interface called `Order`. Which takes us back to the initial `Order` importing that's equal to the new `Order`.

Diogomartf: I think we can use Order from `sdk` now, will update it.

berteotti (approved):

PR #31, opened 2023-07-13
feat[STK-33]: create stack frequency section

berteotti:
Create Stack Frequency section. This PRs add new libs to improve date formats, add popover effect and add calendar components.

Screenshot 2023-07-31 at 11 29 51


packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -3,12 +3,17 @@
...
+import { TokenPicker, DatePicker } from "@/components";
 
 export const Stackbox = () => {
   const [isConfirmStackOpen, setConfirmStackIsOpen] = useState(false);
   const [isTokenPickerOpen, setTokenPickerIsOpen] = useState(false);
 
+  const [frequency, setFrequency] = useState<string>("0");
+
+  const [startDateTime, setStartDateTime] = useState<Date>(new Date());
+  const [endDateTime, setEndDateTime] = useState<Date>(new Date());
```
Diogomartf, line 15: Can we set the default end time date 1 month from now?

I think it would be nice to already be in the future, one month seems nice.

something like this should work (from a quick search)
```suggestion
  const [endDateTime, setEndDateTime] = useState<Date(new Date().setMonth(new Date().getMonth() + 1);
```

berteotti: This didn't work but applied another solution. Thanls!

Diogomartf:
image

**Design feedback**
- on the selected date button the ring should be "primary-200" (or 300)
- set date and time button should be primary action so it should be primary color, it's the main CTA  (if not I forgot to click to save)
- remove @ on the date

Diogomartf:
https://github.com/SwaprHQ/stackly-ui/assets/5664434/6f26a1e3-175f-4fb7-b9d9-0a9193452491


Isn't this behavior odd?

packages/app/app/layout.tsx
```diff
@@ -7,6 +7,12 @@ export const metadata = {
   title: "Stackly | Stack crypto over time.",
   description:
     "Stackly is a simple, non-custodial tool that uses the CoW protocol to place recurring swaps based on DCA..",
+  viewport: {
+    width: "device-width",
+    initialScale: 1,
+    maximumScale: 1,
+    userScalable: 0,
+  },
```
Diogomartf, line 15: why?

packages/app/components/DatePicker.tsx
```diff
@@ -0,0 +1,170 @@
...
+import { format, isToday, isTomorrow } from "date-fns";
+
+import { Button, Calendar, Icon } from "@/ui";
+import { Popover } from "@headlessui/react";
+
+import { Dispatch, Ref, SetStateAction, useEffect, useState } from "react";
+import { usePopper } from "react-popper";
+import { twMerge } from "tailwind-merge";
+
+interface IDatePicker {
```
Diogomartf, line 12: When interface is the props the component receives I'm using `DatePickerProps` naming convention. 

```suggestion
interface DatePickerProps {
```

Don't know if we want to change it to this convention, but think we should stick with one.

packages/app/components/DatePicker.tsx
```diff
@@ -0,0 +1,170 @@
...
+  return (
+    <Popover>
+      {({ open }) => (
+        <>
+          <Popover.Button
+            ref={setReferenceElement as Ref<HTMLButtonElement>}
+            className={twMerge(
+              "flex justify-between items-center focus:border-0",
+              className
+            )} // text-em-low if default dates
```
Diogomartf, line 65:
```suggestion
            )}
```

packages/app/components/DatePicker.tsx
```diff
@@ -0,0 +1,170 @@
...
+                        if (!newDate) return;
+                        newDate.setHours(currentDate.getHours());
+                        newDate.setMinutes(currentDate.getMinutes());
+                        setCurrentDate(newDate);
+                      }}
+                      initialFocus
+                      className=""
+                      fromDate={fromDate || new Date()}
+                    />
+                    <div className="flex flex-col space-y-2 p-4 pt-2">
```
Diogomartf, line 95:
```suggestion
                    <div className="flex flex-col space-y-2 p-4">
```

We have two paddings here, I think pt-2 is not doing anything we can remove it

packages/app/components/DatePicker.tsx
```diff
@@ -0,0 +1,170 @@
...
+                              type="number"
+                              pattern="[0-9]*"
+                              value={hours}
+                              onKeyDown={(evt) =>
+                                ["e", "E", "+", "-"].includes(evt.key) &&
+                                evt.preventDefault()
+                              }
+                              onChange={(event) => {
+                                const value = event.target.value;
+                                const numberValue = Number(value);
```
Diogomartf, line 111: Don't like much of this naming, but.. have no suggestions either

packages/app/components/DatePicker.tsx
```diff
@@ -0,0 +1,170 @@
...
+                              type="number"
+                              pattern="[0-9]*"
+                              value={minutes}
+                              onKeyDown={(evt) =>
+                                ["e", "E", "+", "-"].includes(evt.key) &&
+                                evt.preventDefault()
+                              }
+                              onChange={(event) => {
+                                const value = event.target.value;
+                                const numberValue = Number(value);
```
Diogomartf, line 135: same here

packages/app/components/DatePicker.tsx
```diff
@@ -0,0 +1,170 @@
...
+                                if (
+                                  numberValue >= 0 &&
+                                  numberValue < 24 &&
```
Diogomartf, line 114:
```suggestion
                                const hours = Number(value);
                                if (
                                  hours >= 0 &&
                                  hours < 24 &&
```

`numberOfHours` would also work

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -13,6 +18,28 @@ interface SelectTokenButtonProps {
...
+
+const endDateByFrequency: Record<string, number> = {
+  [HOUR_OPTION]: new Date().setDate(new Date().getDate() + 2),
+  [DAY_OPTION]: new Date().setMonth(new Date().getMonth() + 1),
+  [WEEK_OPTION]: new Date().setMonth(new Date().getMonth() + 3),
+  [MONTH_OPTION]: new Date().setFullYear(new Date().getFullYear() + 1),
+};
+const startDateTimeTimestamp = new Date().setMinutes(
+  new Date().getMinutes() + 30
+);
```
Diogomartf, line 41: this logic is okay here for now, but It would be great to have a place for this logic, maybe utils or create a model.

packages/app/app/layout.tsx
berteotti: We need this to prevent zoom in on mobile when clicking on inputs on iOS

packages/app/components/DatePicker.tsx
berteotti: Both work to me, let's use `DatePickerProps`

packages/app/app/layout.tsx
Diogomartf: I think that only happens on `iOS` not a big deal. Don't know if  has implications in other things, but let's keep it for now.

Diogomartf (dismissed):

Diogomartf (approved):

PR #32, opened 2023-07-13
[STK-89][FEAT] - TokenPicker token select functionality

ElRodrigote:
### Description

* Assets import/export refactor
* Implemented token selection in the `TokenPicker` to be displayed in `Stackbox`

image

### Preview
https://feat-stk-89-tokenpicker-toke.stackly-ui.pages.dev/

### How to test the changes

1) Pull this branch
2) Run the project locally
3) Go to Stackly landing page, click any "Select token" button and wait for the modal to pop-up
4) The TokenPicker view should select a token, close the modal and clear the `SearchBar` only if it has a written query on any:
4.1)`TokenList` interaction
4.2)`CommonTokens` interaction


packages/app/components/navbar/Logo.tsx
```diff
@@ -1,11 +1,12 @@
-import StacklyLogoIcon from "../../public/assets/images/stackly-logo-icon.svg";
-import StacklyLogoName from "../../public/assets/images/stackly-logo-name.svg";
+"use client";
+
+import { StacklyLogoIcon, StacklyLogoNameImg } from "@/public/assets";
```
Diogomartf, line 4: good 👍

packages/app/components/stackbox/StackTransactionLoading.tsx
```diff
@@ -1,15 +1,15 @@
-import DotsSpinner from "public/assets/images/dots-spinner.svg";
+import { DotsSpinnerImg } from "@/public/assets";
```
Diogomartf, line 1: good change 👍

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -91,11 +115,42 @@ export const Stackbox = () => {
...
+    <div
+      className={cx("flex flex-col space-y-2", {
+        "items-end": !isTokenFrom
+      })}
+    >
+      <BodyText className="text-em-low">{label}</BodyText>
+      {token ? (
+        <ChipButton id={token.address} onClick={handleButtonClick} size="sm">
+          <TokenIcon token={token} />
+          <p className="font-semibold ml-1.5">{token.symbol}</p>
```
Diogomartf, line 142: I think we should use our `BodyText` component.

```suggestion
<BodyText className="ml-1.5">{token.symbol}</BodyText>
```

packages/app/components/token-picker/TokenPicker.tsx
```diff
@@ -118,32 +130,31 @@ const SearchBar = forwardRef<HTMLInputElement, SearchBarProps>(
...
           id={token.address}
           key={token.address}
           name={token.name}
-          onChange={() => {}}
-          value={token.name}
+          onClick={() => onTokenSelect(token)}
+          size="sm"
         >
           <TokenIcon token={token} />
           <p className="font-semibold ml-1.5">{token.symbol}</p>
```
Diogomartf, line 146: Same here let's use `BodyText` component.

```suggestion
<BodyText className="ml-1.5">{token.symbol}</BodyText>
```

Diogomartf: Isn't this the same component as above?

Maybe we should create `ChipTokenButton` and have only one of this. Only one place to change.

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -91,11 +115,42 @@ export const Stackbox = () => {
...
       <ConfirmStackModal
         isOpen={isConfirmStackOpen}
         closeAction={closeConfirmStack}
       />
+    </>
+  );
+};
+
+const SelectTokenChip = ({ label, onClick, token }: SelectTokenChipProps) => {
+  const isTokenFrom = label === TOKEN_FROM_LABEL;
```
Diogomartf, line 129: I think I would like something more like this:

```suggestion
  const isTokenFrom = `label.toLowerCase().includes("deposit")`;
```

And then we could remove the constants: `TOKEN_FROM_LABEL` and `TOKEN_TO_LABEL`

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -1,42 +1,66 @@
...
+} from "@/ui";
 import { useRef, useState } from "react";
 import { ConfirmStackModal } from "./ConfirmStackModal";
 import { TokenPicker } from "@/components/token-picker/TokenPicker";
+import { TokenFromTokenlist } from "@/models/token";
+import { TokenIcon } from "../TokenIcon";
+import { cx } from "class-variance-authority";
+
+const TOKEN_FROM_LABEL = "Deposit from";
+const TOKEN_TO_LABEL = "To receive";
```
Diogomartf, line 19: related: https://github.com/SwaprHQ/stackly-ui/pull/32/files#r1263009083

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -1,42 +1,66 @@
...
+  const [isPickingTokenFrom, setIsPickingTokenFrom] = useState<boolean>(false);
+  const [tokenFrom, setTokenFrom] = useState<TokenFromTokenlist>();
+  const [tokenTo, setTokenTo] = useState<TokenFromTokenlist>();
 
   const openConfirmStack = () => setConfirmStackIsOpen(true);
   const closeConfirmStack = () => setConfirmStackIsOpen(false);
 
-  const openTokenPicker = () => setTokenPickerIsOpen(true);
+  const openTokenPicker = (isTokenFrom = true) => {
+    setIsPickingTokenFrom(Boolean(isTokenFrom));
```
Diogomartf, line 38: why do we need to cast it as `Boolean`?

Shouldn't it always come a boolean?

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -91,11 +115,42 @@ export const Stackbox = () => {
         closeAction={closeTokenPicker}
         initialFocusRef={searchTokenBarRef}
         isOpen={isTokenPickerOpen}
+        onTokenSelect={isPickingTokenFrom ? setTokenFrom : setTokenTo}
```
Diogomartf, line 118: I would prefer this to be function, as it a condition I would better have it above next to other logic.

Something like this:
```js
const selectToken => {isPickingTokenFrom ? setTokenFrom : setTokenTo)
```

packages/app/components/stackbox/Stackbox.tsx
ElRodrigote: Ahhh agree, seen this a few times and was In doubt but I'd replace every `<p>` tag with our `BodyText` component lol. Will change it here!

packages/app/components/token-picker/TokenPicker.tsx
ElRodrigote: It's not exactly the same, I'll end up deleting the other ChipTokenButton as we're not needing it and we can solve the approach with either `ChipButton` or plain `Button` :D

packages/app/components/stackbox/Stackbox.tsx
ElRodrigote: I think we should come to an agreement regarding variable creation. 
In a previous comment, I've created some constant variables for text reusability purposes (for `SelectTokenButton` labels and checks), that we suggested to remove to achieve (I guess) code simplicity and a more lean codebase.

But now I see we're creating a lot of variables for simple stuff like:
```ts
  const closeConfirmStack = () => setConfirmStackIsOpen(false);
  const closeTokenPicker = () => setTokenPickerIsOpen(false);
  const openConfirmStack = () => setConfirmStackIsOpen(true);
  const selectToken = isPickingTokenFrom ? setTokenFrom : setTokenTo;
  ```
  
  I don't mind either way on this matter TBH, but I'd suggest taking a decision for code normalization purposes

Diogomartf: Yeah, we should normalize that.

 We can remove the variables and do it directly on the component props (on simple logic like this).

packages/app/components/index.ts
Diogomartf, line 1: Curious, why this approach is better ?

PR #33, opened 2023-07-14
feat: get chain and address from wallet and improve stacks calculations

Diogomartf:
**Summary**
- convert /stacks page into client component
- `useNetwork` to get chain
- `useAccount` to get connect account hook
- add loading states
- remove "use client" from place where is not needed
- move `not-found `page into root


Diogomartf:
Updated to `next@13.4.10` seems like it's wokring good!

@berteotti can we merge this?

packages/app/app/not-found.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const discordLink = process.env.DISCORD_URL ? process.env.DISCORD_URL : "#";
+
+  return (
+    <div className="flex flex-col items-center justify-center max-w-xl mx-auto my-32 space-y-7">
+      <StacksImg />
+      <HeadingText size={3}>Page not found</HeadingText>
+      <HeadingText className="text-center text-em-med">
+        {`We're sorry, but the page you are looking for cannot be found. Please
+        check the URL and try again. If you believe this is an error, please
+        contact us on our`}{" "}
```
berteotti, line 17: This can be simple text, right?

packages/app/app/stacks/stacks-orders.tsx
```diff
@@ -0,0 +1,116 @@
...
+    fetchStackOrdersData();
+  }, [chainId, currentOrders]);
+
+  useEffect(() => {
+    if (currentStackOrders?.length) {
+      setCompletedOrders(filterCompletedOrders(currentStackOrders));
+      setActiveOrders(filterActiveOrders(currentStackOrders));
+      setLoading(false);
+    }
+  }, [currentStackOrders]);
```
berteotti, line 54: If I don't create any `currentStackOrders` it will create an infinite loading state.

packages/app/app/stacks/stacks-orders.tsx
```diff
@@ -0,0 +1,116 @@
...
+    fetchOrdersData();
+  }, [address, chainId]);
+
+  useEffect(() => {
+    async function fetchStackOrdersData() {
+      const stackOrders = await getStackOrders(chainId, currentOrders);
+      if (stackOrders) setCurrentStackOrders(stackOrders);
+    }
+    fetchStackOrdersData();
+  }, [chainId, currentOrders]);
```
berteotti, line 46: I don't see a reason for saving the `currentOrders`. Maybe we can join this 2 fetches.

1. Fetch Orders
2. If orders, then fetch stackOrders
3. If no orders, stop loading
4. If stackOrders, then filter orders, set completed and active orders and stop loading
5. If no stackOrders, stop loading

In this case we just need to save active and completed orders, so less 2 `useState`.

packages/app/components/StackedTokenLogoPair.tsx
```diff
@@ -1,3 +1,5 @@
+// @ts-nocheck
+
```
berteotti, line 2: Why do we need this?

packages/app/components/stack-modal/StackModal.tsx
```diff
@@ -1,11 +1,10 @@
+// @ts-nocheck
```
berteotti, line 1: Also, why we need this?

packages/app/components/stack-modal/StackOrdersTable.tsx
```diff
@@ -79,18 +79,22 @@ export const StackOrdersTable = ({ stackOrder }: StackOrderProps) => {
...
-  const formatAmount = (amount: string) =>
-    convertedAmount(amount, 18).toFixed(4);
+const TableCowBody = ({
+  stackOrder,
+  cowOrders,
+}: {
+  stackOrder: StackOrder;
+  cowOrders: CowOrder[];
+}) => {
+  const chainId = 100; // @todo use context or useNetwork hook
```
berteotti, line 89: This seems to be missing

berteotti: Also, where is `chainId` being used?

packages/app/components/stack-modal/StackProgress.tsx
```diff
@@ -1,12 +1,14 @@
-import {
-  fundsAmountWithToken,
-  totalOrdersDone,
-  totalFundsUsed
-} from "@/models/order";
+// @ts-nocheck
```
berteotti, line 1: Why are we using this?

packages/app/models/cow-order/cow-order.ts
```diff
@@ -1,15 +1,15 @@
-type ChainId = 1 | 100;
+import { ChainId } from "@stackly/sdk";
```
berteotti, line 1: There are others places where we are using `1 | 100`. Can you fixed those?

packages/app/app/not-found.tsx
Diogomartf: typescript doesn't like apostrophe, so we have type error on _We're_, that's why the braces

packages/app/components/stack-modal/StackProgress.tsx
Diogomartf: same as above

packages/app/app/stacks/stacks-orders.tsx
Diogomartf: Yeah, probably we should do this. I'll fix that.

packages/app/app/stacks/stacks-orders.tsx
Diogomartf: I'll fix that, I think related with bellow.

packages/app/app/stacks/stacks-orders.tsx
```diff
@@ -0,0 +1,112 @@
...
+  chainId: ChainId;
+}
+
+export const StackOrders = ({ chainId, address }: StackOrdersProps) => {
+  const [loading, setLoading] = useState(true);
+  const [completedOrders, setCompletedOrders] = useState<StackOrder[]>([]);
+  const [activeOrders, setActiveOrders] = useState<StackOrder[]>([]);
+  const [currentStackOrders, setCurrentStackOrders] = useState<StackOrder[]>(
+    []
+  );
```
berteotti, line 29: There is no necessity for this currentStackOrders. We can populate active and complete orders directly and there is no need for an extra `useEffect`. To check the empty state we just need to check if both active and complete orders are empty.

packages/app/components/stack-modal/StackOrdersTable.tsx
Diogomartf: `cowExplorerUrl` function bellow

Diogomartf: we're hardcoding this, we have to remove ethereum from the networks for now.

Diogomartf: We should do it on other PR.

Created a task for that
https://linear.app/swaprdev/issue/STK-101/remove-ethereum-chain-from-networks-and-remove-hardcoded-chainids

packages/app/app/not-found.tsx
Diogomartf: it should. Fixed!

packages/app/models/cow-order/cow-order.ts
Diogomartf: Didn't find any, other were hardcoded that's related to this: 

https://github.com/SwaprHQ/stackly-ui/pull/33#discussion_r1280881022

packages/app/app/stacks/stacks-orders.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  useEffect(() => {
+    function fetchData() {
+      getOrders(chainId, address.toLowerCase())
+        .then(async (orders) => {
+          if (!orders) return;
+
+          const stackOrders = await getStackOrders(chainId, orders);
+          if (stackOrders.length > 0) setCurrentStackOrders(stackOrders);
```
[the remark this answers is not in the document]
Diogomartf: `if (!orders) return;` this check does that.

packages/app/app/stacks/stacks-orders.tsx
Diogomartf: we can have just currentStackOrders and not the active/complete orders on state

berteotti: Nice, that also works

berteotti (approved):
LGTM

packages/app/app/stacks/stacks-orders.tsx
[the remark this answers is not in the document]
Diogomartf: okay I'll update to:

`if (!orders || orders.length === 0) return;`

PR #11, opened 2023-05-11
feat: add meta open graph and twitter images

Diogomartf:
This PRs adds open graph and twitter meta images.

Following Next js file convention: https://nextjs.org/docs/app/api-reference/file-conventions/metadata/opengraph-image

PR #34, opened 2023-07-25
[STK-90][REFACTOR] - Modal component dividers, DOM tree improvements, docs enhancement, Nextjs errors fixing

ElRodrigote:
### Description

* Dividers are now conditional styles for Modal sub-components. Handling them with `divide-y` from Tailwind is a no-go for now as we don't have dividers consistency across current modals.
* Created a new React Context to handle all the dapp Modals.
* Optimized dapp current Contexts by stabilizing their referenced values

  #### Minor stuff:
* Remove unneeded/unused files like duplicated icons and the whole `packages/app/app/ui` folder.
* Add `Troubleshooting` and `Code Guidelines` to the README, with some examples.
* General improvements in the DOM tree, removing unneeded elements, updating imports, and module renaming.
* Removed a few Nextjs warnings regarding hydration issues.

### Preview
https://refactor-stk-90-modal-compon.stackly-ui.pages.dev/

### How to test the changes

1) Pull this branch
2) Run the project locally
3) Go to Stackly landing page, both app behavior and app look&feel should be the same, maybe a bit faster


packages/app/app/icon.ico
Diogomartf, line 1: I think we do want the `favicon.ico`, it's the favicon, so the name is correct.

What's the reasoning on this?

Diogomartf:
I thin this PR has a lot more changes than `Modal`, so the current title doesn't represent the changes. I would ask you to rename it for something more descriptive.

ElRodrigote:
> This PR has a lot more changes than `Modal` and dividers, so the current title doesn't represent the changes. I would ask you to rename it for something more descriptive.
> 
> Also, next time, let's break it into smaller PRs. It makes it much easier to review the changes.

Yes, you're 100% right. Sorry for mixing a lot of different things in here :c

Will bear this in mind for the future!

packages/app/app/stacks/page.tsx
```diff
@@ -37,48 +27,7 @@ export default async function Page() {
...
-              {hasCompletedOrders ? (
-                <StacksTable stackOrders={completedOrders} />
-              ) : (
-                <EmptyStacks text="No completed stacks" />
-              )}
-            </Tab.Panel>
-          </Tab.Panels>
-        </Tab.Group>
-      </div>
+      <StackTabs stackOrders={stackOrders} />
```
Diogomartf, line 30: I would prefer to have the tabs exposed and not on a component.
I think that logic fits very well on this page. 

Also don't think that `StacksModal` belongs there. We might used it elsewhere, not related to the tabs.

I would revert this changes.

packages/app/components/stack-tabs/StackTabs.tsx
Diogomartf, line 1: Also don't think that `StacksTable` belongs here. We might used it elsewhere, not related to the tabs.

Diogomartf: Same as above, I think this is a it's own component. Not related to tabs.

packages/app/contexts/ModalContext.tsx
```diff
@@ -0,0 +1,48 @@
...
+});
+
+export interface ModalContextProviderProps {
+  children: ReactNode;
+}
+
+export const ModalContextProvider = ({
+  children,
+}: ModalContextProviderProps) => {
+  const [modalToOpen, setModalToOpen] = useState<ModalId | null>(null);
```
Diogomartf, line 30: related to comment above:

```suggestion
  const [modalToOpen, setModalOpenId] = useState<ModalId | null>(null);
```

packages/app/app/stacks/page.tsx
ElRodrigote: The thing is the Stack tabs use Headless UI, and Headless UI is a client-side only package. Using Headless-UI in a server component will trigger errors for hydration issues. And I wouldn't transform Nextjs routing pages (so far the Home page with the stackbox and the Stacks page with the stacks tabs) into client components.

To remove these warnings I abstracted the logic dependant on Headless packages into their own components.

More info: https://github.com/tailwindlabs/headlessui/issues/2021

We can revert this change and keep the warnings or transform this page into a client-side component, but IMHO these two options are sub-optimal :c

packages/app/app/icon.ico
Diogomartf: From [Next docs](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/app-icons)

> Use an image file to set an app icon by placing a favicon, icon, or apple-icon image file within your /app directory. The favicon image can only be located in the top level of app/.

They pickup the `favicon.ico`, `icon` and `apple-icon`on app folder.
`

Might be related, there was a bug that they fixed with new next version:

 https://github.com/SwaprHQ/stackly-ui/pull/11

(also if you could review this PR 👀 )

packages/app/components/stack-tabs/StackTabs.tsx
Diogomartf: sorry, `StackOrdersTable`

packages/app/app/icon.ico
ElRodrigote: This is so weird, when I tried to follow that naming convention from the docs it was not picking up the favicon, it was picking up the `icon.svg`. Will try it again and check what's happening

packages/app/app/stacks/page.tsx
Diogomartf: I'm okay with this change if it improves the hydration issues and DEX.

But let's rename it to `StackOrders`. 

Having tabs or not is not as important on the name, I want to know it's StackOrders and then if I go to the component I can see we filter by active/complete and we use tabs.

are you okay with this change?

Also would me keep `stack-modal` folder and related components there.

ElRodrigote: Yeah, I'm ok with that. 

Quick question, want me to move the newly named `StackOrders` into `@/components/StackOrders` as a standalone component?

packages/app/components/stack-tabs/StackTabs.tsx
ElRodrigote: I see, related to the previous comment on renaming the folder. So I might rename the folder back to `stack-modal` and then move `StackOrders` as a standalone component outside this renamed folder, do you agree?

packages/app/app/icon.ico
Diogomartf: I think will pick up once we merge the PR above. So I would revert the change and keep it as it is.

packages/app/app/stacks/page.tsx
Diogomartf: Yes, I think it can be like that.

packages/app/components/stack-tabs/StackTabs.tsx
Diogomartf: Yes, I think it works.

packages/app/app/ui/page.tsx
Diogomartf, line 1: why remove this page?

It's handy to see the UI components.

packages/app/app/stacks/page.tsx
Diogomartf: Oh, already splited that logic in my PR, I have [StackOrders component](https://github.com/SwaprHQ/stackly-ui/pull/33/files#diff-56341ad50f06b78bbb4e3ed75ca24713793fa0c41972c0980e91e2b8e6bcf688R22-R54).

Maybe I can move it to other folder, and you revert the changes here.

Diogomartf: Also copied your logic on `activeOrders/completedOrders`.

packages/app/app/icon.ico
ElRodrigote: You were right! I've just tried and it picks up that naming convention. Maybe I got some cached data that messed up the time I tried it.

Thanks!

packages/app/app/stacks/page.tsx
ElRodrigote: Reverting these changes so we can have them from your work :D

Regarding if I keep getting errors:

For this file I get either a ["no async server components"](https://nextjs.org/docs/messages/no-async-client-component) if I keep the `"use client"` directive in this file.

If I move the `"use client"` directive out of this page component and into the `StacksTable` component, I get a crashing error while building the app, because we're trying to use `Tabs` from Headless UI (client-only lib) in the page component.
image

I could solve these errors by abstracting the component that implements `Tabs` into its own component having the `"use client"` directive in there.

packages/app/app/ui/page.tsx
ElRodrigote: My bad, thought we were not using this page and removed it (because I didn't see a way to navigate to it, my bad).

I'd move this whole page into a Storybook or something like that in the future tho. Will restore it!

Diogomartf: No prob. We're using /ui for listening components. 

Yeah, we can move to Storybook but for now I think this is enough.

packages/app/ui/modal/ModalContent.tsx
```diff
@@ -1,12 +1,28 @@
...
+export function ModalContent({
+  children,
+  className,
+  withDivider = false,
+}: ModalContentProps) {
+  return (
+    <div
+      className={twMerge(
+        cx("px-4 mt-5", className, {
+          "border-b border-surface-50": withDivider,
```
berteotti, line 21: I don't see any use case for this for now. Should we add this here?

packages/app/ui/modal/ModalHeader.tsx
```diff
@@ -1,13 +1,24 @@
...
+  withDivider = false,
+}: ModalHeaderProps) {
   return (
     <Dialog.Title
       as="div"
-      className="flex items-center justify-between w-full border-b border-surface-50 px-4 py-3"
+      className={cx("flex items-center justify-between w-full px-4 py-3", {
+        "border-b border-surface-50": withDivider,
+      })}
     >
```
berteotti, line 22: Are we defining some `ModalHeader` without divider? If not, I don't see why we should define this

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -71,100 +69,98 @@ export const Stackbox = () => {
...
-            />
-          </div>
-          <div className="py-2">
-            <input
-              type="number"
-              placeholder="0.0"
-              className="w-full py-3 text-4xl font-semibold outline-none"
-            />
-          </div>
+    <div className="max-w-lg mx-auto my-24 bg-white shadow-2xl rounded-2xl">
```
berteotti, line 72: I don't think this style belongs to the component. This style is related to the page. If we want to have a Stackbox inside a Modal, for example, we can't have this style. This mental exercise helps me defining where each style belongs (inside the component or inside the page). But what do you think? :)

Talking about `max-w-lg mx-auto my-24`.

berteotti (commented):
PR is looking good, left a few comments.

packages/app/components/stackbox/Stackbox.tsx
ElRodrigote: OIC I thought the Stackbox WAS kinda the page content itself. I'll refactor this one, thanks for the heads up!

packages/app/ui/modal/ModalContent.tsx
ElRodrigote: Will remove it, it was intended to be used for the Toast component and enhance future modals that may use a divider between the content and the footer, but now is not needed as the Toast component was created with another approach.

packages/app/ui/modal/ModalHeader.tsx
ElRodrigote: Will remove it, it was intended to be used for the Toast component and enhance future modals that may use a divider between the content and the footer, but now is not needed as the Toast component was created with another approach.

Diogomartf:
It would also be nice to have different examples of toast on `/ui`

I think it can have a section like `Modal`

berteotti (approved):
Looks good to me!

Diogomartf (dismissed):

Diogomartf (approved):

PR #35, opened 2023-07-28
[STK-75][FEAT] - Add Toast component 

ElRodrigote:
### Description

* Updated Modal logic to opt-out the background opacity
* Created the Toast component
* Updated UI page where we display Stackly components

### Preview
image
https://feat-stk-75-add-toast-compon.stackly-ui.pages.dev/ui

### How to test the changes

1) Pull this branch
2) Run the project locally
3) Go to the Stackly UI components page (`https://feat-stk-75-add-toast-compon.stackly-ui.pages.dev//ui`) and check for the Modals section, you will find a "Toast" button.


packages/app/ui/toast/Toast.tsx
```diff
@@ -0,0 +1,110 @@
...
+    return () => {
+      clearInterval(interval);
+    };
+  }, [autoClose, autoCloseTime, closeAction, isOpen]);
+
+  const hasSeverity = Boolean(severity);
+
+  return (
+    <Modal
+      className={cx("py-6 pl-6 pr-4 !w-fit", toastStyles({ position }))}
```
berteotti, line 55: `"py-6 pl-6 pr-4 !w-fit"` belong to default toastStyles, right?

ElRodrigote: Yes! Exactly :D

OIC maybe I could move them into `toastStyles`

berteotti (dismissed):
LGTM, good job Rorry!

Diogomartf:
Design wise there are two things I don't think are good:
- Clicking outside of the Dialog will close it
- Not able to scroll page without dealing with the dialog.

Ideally we should have notifications appear and leave without having to interact with them.

If this issues are hard to come around, I'm okay with shipping this for MVP (as it's not that bad), we can improve it in future iterations.

I like the fact that we're reusing a component we already have.

ElRodrigote:
> Design wise there are two things I don't think are good:
> 
> * Clicking outside of the Dialog will close it
> * Not able to scroll page without dealing with the dialog.
> 
> Ideally we should have notifications appear and leave without having to interact with them.
> 
> If this issues are hard to come around, I'm okay with shipping this for MVP (as it's not that bad), we can improve it in future iterations.
> 
> I like the fact that we're reusing a component we already have.

I can come around with a solution around this but it won't be reusing our Modal component, as the Modal component is a Dialog applying backgrounds. Fundamentally speaking, a Modal and a Toast have different purposes. At first, I was trying to design a component that could bear a list of notifications but I thought it was going way over the top for our needs.

If you agree I can try and create something in between: lay down the groundwork for a Toast component not using our Stackly UI Modals, being able to keep interacting with the dApp, and default close the Toast.

WDYT?

Diogomartf:
> > Design wise there are two things I don't think are good:
> > 
> > * Clicking outside of the Dialog will close it
> > * Not able to scroll page without dealing with the dialog.
> > 
> > Ideally we should have notifications appear and leave without having to interact with them.
> > If this issues are hard to come around, I'm okay with shipping this for MVP (as it's not that bad), we can improve it in future iterations.
> > I like the fact that we're reusing a component we already have.
> 
> I can come around with a solution around this but it won't be reusing our Modal component, as the Modal component is a Dialog applying backgrounds. Fundamentally speaking, a Modal and a Toast have different purposes. At first, I was trying to design a component that could bear a list of notifications but I thought it was going way over the top for our needs.
> 
> If you agree I can try and create something in between: lay down the groundwork for a Toast component not using our Stackly UI Modals, being able to keep interacting with the dApp, and default close the Toast.
> 
> WDYT?

I think we can ship the Toast component as it is. 

You can create a task to build a proper Toast component that handles those things I pointed nicely. 

And do it when we have some time after the MVP as it's not a priority.

Diogomartf (approved):

berteotti (approved):

PR #36, opened 2023-07-31
fix: deploy on fleek

berteotti:
Fixes deploy on fleek and IPFS.


Diogomartf (approved):

Diogomartf:
I would arg this is not a fix but an update to allow deploy on fleek.

berteotti:
> I would arg this is not a fix but an update to allow deploy on fleek.

I can call it `build` as it is build related

PR #38, opened 2023-08-02
feat: create StackTokenPair

berteotti:
Screenshot 2023-08-02 at 11 41 59

Creates StackTokenPair section. Formatted balance isn't consistent throughout the code base.


Diogomartf:
While testing in Brave got Application error.

I select a token on the token picker

image

Looking at the console it looks like:

`RangeError: maximumSignificantDigits value is out of range.`
`    at Number.toLocaleString 

image

Diogomartf:
I think we should clean the input when changing tokens, 

what do you think?

4899bfc53935811d1e727a2438109f09

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -70,6 +86,14 @@ export const Stackbox = () => {
     setEndDateTime(new Date(endDateByFrequency[frequency]));
   }, [frequency]);
 
+  const formattedBalance = (balanceData: NonNullable<typeof balance>) =>
```
Diogomartf, line 89: this function could be on `utils/token` file

berteotti:
> Would be nice that the size of decimals was the same as balance on bottom right.
> 
> like this `8,0994` instead of `8,099426770481235208`
> 
> image

I think it's important to let the user define the granularity of what he is spending. If we hide tokens he can't know if he is using max balance or just a portion of it.

berteotti:
> I think we should clean the input when changing tokens,
> 
> what do you think?
> 
> 4899bfc53935811d1e727a2438109f09

I think we should let it for now. If the user inputs a custom value and it changes token, I think we probably wants to keep that value. Also thought about this and the behaviour on other apps is to keep the value.

packages/app/components/stackbox/Stackbox.tsx
berteotti: I think we should uniformize all this formates in the app. And add this formatting to a specific file. If you agree will create a task to do this. But we can keep this function inside Stackbox for this PR.

ElRodrigote (approved):

packages/app/components/stackbox/Stackbox.tsx
ElRodrigote: Agree on standardizing formats throughout the app but I agree with Diogo to abstract this function into a `utils` file, because, in my experience, these functions tend to grow and change their signature as the app needs to grow. At some point, I'm sure we will abstract this function, sooner or later.

I don't mind handling that abstraction in a different PR as long as we handle it, either here or in another task.

Diogomartf:
> > Would be nice that the size of decimals was the same as balance on bottom right.
> > like this `8,0994` instead of `8,099426770481235208`
> > image
> 
> I think it's important to let the user define the granularity of what he is spending. If we hide tokens he can't know if he is using max balance or just a portion of it.

Just use the same as balance (as we also choose to not show all the decimals of the balance) and use only that portion not the whole decimals. That's just too many numbers.

We let the user go to that kind of granularity if he wants.

berteotti:
> > > Would be nice that the size of decimals was the same as balance on bottom right.
> > > like this `8,0994` instead of `8,099426770481235208`
> > > image
> > 
> > 
> > I think it's important to let the user define the granularity of what he is spending. If we hide tokens he can't know if he is using max balance or just a portion of it.
> 
> Just use the same as balance (as we also choose to not show all the decimals of the balance) and use only that portion not the whole decimals. That's just too many numbers.
> 
> We let the user go to that kind of granularity if he wants.

How can we let the user define the granularity if he doesn't know the entire balance? If the user clicks Max we have to assume he wants to spend the entire token amount, not just an approximation of that value. What do you think?

Diogomartf:
> > > > Would be nice that the size of decimals was the same as balance on bottom right.
> > > > like this `8,0994` instead of `8,099426770481235208`
> > > > image
> > > 
> > > 
> > > I think it's important to let the user define the granularity of what he is spending. If we hide tokens he can't know if he is using max balance or just a portion of it.
> > 
> > 
> > Just use the same as balance (as we also choose to not show all the decimals of the balance) and use only that portion not the whole decimals. That's just too many numbers.
> > We let the user go to that kind of granularity if he wants.
> 
> How can we let the user define the granularity if he doesn't know the entire balance? If the user clicks Max we have to assume he wants to spend the entire token amount, not just an approximation of that value. What do you think?

I would prefer to not be shown so many decimals. It's so small that only adds noise. 

But I'll let you decide on it now. We can change it in the future if we want (add/remove decimals).

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -70,6 +86,28 @@ export const Stackbox = () => {
     setEndDateTime(new Date(endDateByFrequency[frequency]));
   }, [frequency]);
 
+  const formattedBalance = (balanceData: NonNullable<typeof balance>) => {
+    const SIGNIFICANT_DIGITS = 5;
+
+    /* parseFloat as some number size limitations so this is a temporary solution */
```
Diogomartf, line 92:
```suggestion
    /* parseFloat has some number size limitations so this is a temporary solution */
```

what are the limitations? i
s the fix so hard or so necessary that we have to keep a comment?

Diogomartf (dismissed):

packages/app/components/stackbox/Stackbox.tsx
berteotti: Description is clearer now.

Diogomartf (approved):

PR #39, opened 2023-08-03
feat[STK-103]: update subgraph and add address to token entity

berteotti:
- Update subgraph code to the last changes on Stackly repo.
- Add address to token entity.

With this changes, token entity will have an address field.


packages/subgraph/src/mappings/order.ts
```diff
@@ -26,7 +34,25 @@ export function handleDCAOrderInitialized(event: Initialized): void {
...
+  let tryProtocolFee = factory.try_protocolFee();
+  if (!tryProtocolFee.reverted) {
+    protocolFee = BigInt.fromI32(tryProtocolFee.value);
+  }
+
   order.amount = orderContract.amount();
+  order.fee = protocolFee;
+  order.feeAmount = order.amount
+    .times(protocolFee)
+    .div(HUNDRED_PERCENT.minus(protocolFee));
```
Diogomartf, line 55: is this related to adding address to `Token`?

Diogomartf:
It seems the title of the PR doesn't match the changes, it would be nice to be updated as well as the description.

packages/subgraph/src/mappings/order.ts
berteotti: There are 2 commits in this PR. One to update the code to the last changes on Stackly repo another to add the address to token entity.

Diogomartf (approved):

packages/subgraph/src/mappings/order.ts
```diff
@@ -26,7 +34,25 @@ export function handleDCAOrderInitialized(event: Initialized): void {
   order.sellToken = createOrReturnTokenEntity(orderContract.sellToken()).id;
   order.buyToken = createOrReturnTokenEntity(orderContract.buyToken()).id;
   order.receiver = orderContract.receiver();
+
+  let protocolFee: BigInt = BigInt.fromI32(0);
+  let orderFactoryAddress: Address = (
+    event.transaction.to !== null
+      ? event.transaction.to
+      : Address.fromString("0x0")
+  )!;
```
ElRodrigote, line 43: Minor, non-blocking, this suggestion should yield the same result:

```suggestion
  let orderFactoryAddress: Address = (
    event.transaction.to ?? Address.fromString("0x0")
  )!;
```

More info: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

packages/subgraph/schema.graphql
```diff
@@ -18,13 +18,15 @@ type OrderFactory @entity {
 # Provides information about an ERC20 token
 #
 type Token @entity {
-  " Token address "
+  "Token address "
```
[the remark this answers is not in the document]
berteotti: We have `id` which is an Token address and have `address` which also is a Token address. So I think there is no way to do it differently :)

packages/subgraph/src/mappings/order.ts
berteotti: Since this is AssemblyScript it doesn't support the operator you are suggesting. But nice suggestion anyway :)

Diogomartf (approved):

PR #42, opened 2023-08-09
feat[STK-34]: create stack confirmation modal

berteotti:
- Create stack confirmation modal
- Stack form error handling


packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -14,30 +14,28 @@ import { TokenFromTokenlist } from "@/models/token";
...
 const frequencyOptions = [
-  { option: HOUR_OPTION, name: "Hour" },
-  { option: DAY_OPTION, name: "Day" },
-  { option: WEEK_OPTION, name: "Week" },
-  { option: MONTH_OPTION, name: "Month" },
+  { option: FREQUENCY_OPTIONS.hour, name: "Hour" },
+  { option: FREQUENCY_OPTIONS.day, name: "Day" },
+  { option: FREQUENCY_OPTIONS.week, name: "Week" },
+  { option: FREQUENCY_OPTIONS.month, name: "Month" },
 ];
```
ElRodrigote, line 30: **Minor, non-blocking, suggestion that can be dismissed**
We may not need this one as we could simply use the `FREQUENCY_OPTIONS`

We could create a `titleCase` function to capitalize the first letter on a string into the utils folder (I'm assuming we would be using this one frequently) and just use the keys from the enum:
At line 220, where we use this variable:
```tsx
<div className="flex space-x-2">
  {Object.keys(FREQUENCY_OPTIONS).map((option) => {
    const isSelected = frequency === option;
    return (
      <RadioButton
        key={option}
        name={option}
        id={option}
        checked={isSelected}
        value={option}
        onChange={(event) =>
          setFrequency(event.target.value as FREQUENCY_OPTIONS)
        }
      >
        <BodyText
          size={2}
          className={cx({ "text-em-med": !isSelected})}
        >
          {titleCase(option)}
        </BodyText>
      </RadioButton>
    );
  })}
</div>
```

The `titleCase` util function could look something like this: 
```ts
export const titleCase = (word: string) =>
  word[0].toUpperCase() + word.slice(1).toLowerCase();
```

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -51,39 +49,68 @@ const balanceOptions = [
...
+      endDateTime.getTime() <= startDateTime.getTime();
+    const tokenAmountIsZero = tokenAmount === "0";
+
+    if (!tokenAmount) {
+      setShowTokenAmountError(true);
+    }
+
+    if (endTimeBeforeStartTime) {
+      setDateTimeError(true);
+    }
```
ElRodrigote, line 99: **Minor, non-blocking, prefer one-liners when possible** 

```suggestion
    if (!tokenAmount) setShowTokenAmountError(true);
    if (endTimeBeforeStartTime) setDateTimeError(true);
```

packages/app/components/stackbox/Stackbox.tsx
berteotti: Personally I think using hardcoded works better.

1. Easier to read and update
2. Less code = less bugs
3. Unnecessary JS running
4. We need to rewrite the logic if having translations or different names.

I think what you suggested  it's very handy, but I personally think it's better to keep it simple. What do you think?

packages/app/components/stackbox/ConfirmStackModal.tsx
```diff
@@ -13,30 +13,76 @@ import {
...
+};
+
+const HOUR_IN_MILISECONDS = 60 * 60 * 1000;
+const DAY_IN_MILISECONDS = 24 * HOUR_IN_MILISECONDS;
+
+const frequencySeconds = {
+  [FREQUENCY_OPTIONS.hour]: HOUR_IN_MILISECONDS,
+  [FREQUENCY_OPTIONS.day]: DAY_IN_MILISECONDS,
+  [FREQUENCY_OPTIONS.week]: 7 * DAY_IN_MILISECONDS,
+  [FREQUENCY_OPTIONS.month]: 30 * DAY_IN_MILISECONDS,
```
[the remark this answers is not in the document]
berteotti: This is being used to calculate the number of orders per stack. So we can estimate how much the user will spend for each order. 

About intervals, they work based on hours at contract level. So 1 month = 24 hours * 30 days = 720 hours. Right now we can't specify custom days to run orders, but that's under the scope for new contracts logic.

packages/app/components/stackbox/ConfirmStackModal.tsx
```diff
@@ -56,37 +102,62 @@ export const ConfirmStackModal = ({ isOpen, closeAction }: ModalBaseProps) => {
...
           <div>
             <TitleText size={2} className="text-center text-em-low">
-              Stacks <span className="text-em-high">62.5 USDC</span> worth of{" "}
-              <span className="text-em-high">WETH</span> every hour
+              Stacks{" "}
+              <span className="text-em-high">
+                {amountPerOrder} {fromToken.symbol}
+              </span>{" "}
+              worth of <span className="text-em-high">{toToken.symbol}</span>{" "}
+              every {frequencyTitles[frequency]}
```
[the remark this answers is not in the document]
berteotti: Works for me!

packages/app/components/stackbox/ConfirmStackModal.tsx
```diff
@@ -13,30 +13,76 @@ import {
...
   };
 
+  const estimatedNumberOfOrders =
+    Math.floor(
+      (endTime.getTime() - startTime.getTime()) / frequencySeconds[frequency]
+    ) + INITAL_ORDER;
+
+  const amountPerOrder = (parseFloat(amount) / estimatedNumberOfOrders).toFixed(
+    2
+  );
```
Diogomartf, line 84: maybe this belongs in stack model file

packages/app/components/stackbox/ConfirmStackModal.tsx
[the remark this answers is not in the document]
Diogomartf: @ElRodrigote suggestion makes sense.

packages/app/components/stackbox/Stackbox.tsx
Diogomartf: +1

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -51,39 +49,68 @@ const balanceOptions = [
...
+
+    if (
+      fromToken &&
+      toToken &&
+      !endTimeBeforeStartTime &&
```
Diogomartf, line 104: _just an `OCD` thing (feel free to ignore)_

Can we move this condition `!endTimeBeforeStartTime`  to be the last thing?

Find it easier to read.

packages/app/models/stack/types.ts
```diff
@@ -0,0 +1,6 @@
+export enum FREQUENCY_OPTIONS {
+  hour = "hour",
+  day = "day",
+  week = "week",
+  month = "month",
+}
```
Diogomartf, line 6: is this a type? 

I see types as typescript things: types, interfaces, props, etc.

 I think it belong to the `Stack` model

packages/app/components/stackbox/ConfirmStackModal.tsx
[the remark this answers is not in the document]
ElRodrigote: Ahhh gotcha, so it's a constraint on the SC, makes a lot of sense to me right now. Thanks for the clarification! Marking this as resolved 👌

packages/app/components/stackbox/Stackbox.tsx
ElRodrigote: I don't have a strong opinion right now TBH. In my head, it made more sense to use the `enum` directly but the current app state feels the same. Right now, the util I suggested would be used in this file only.

Having i18n will imply fairly the same amount of work with either path we take IMHO.

So all in all I'd say if you're not quite convinced, dismiss these suggestions as doing so imply less work now :D

Marking this one as resolved 👌

packages/app/models/stack/types.ts
ElRodrigote: `enum` can be used for types. The main difference I'd point out is if you want to allow the dev to hardcode text values or to use the `enum` as single source of truth.

So these two should yield the same results (type-wise speaking):
`type FrequencyOptions = "hour" | "day" | "week" | "month";`

But the way to pass down these values changes. If you type a value as `FrequencyOptions`, then you can pass down `"hour"` as string and it will pass.

If you type a value as `FREQUENCY_OPTIONS`, then you HAVE to pass down `FREQUENCY_OPTIONS.hour` as the value, because if you try to pass down `"hour"` string value it will fail.

----------------------

I like enums better when it's a specific set of string/number values, as you always use the same source of truth and if you have to update the source used for type, updating the `enum` will impact everywhere where it's being used, whereas updating the `type` will also imply to update all the places where it's being used.

packages/app/components/stackbox/ConfirmStackModal.tsx
berteotti: I like to have this things declared next to where they are being used. If this was being used elsewhere I definitely would declare under stack model file. What do you think?

packages/app/models/stack/types.ts
berteotti: As Rorry said, it can be used for types (and I defined this way to use it as a type), so I would leave it here.

packages/app/components/stackbox/Stackbox.tsx
berteotti: Fixed in c9a830dae34db6f75e49a63d9b2c0a972db77ef6

packages/app/components/stackbox/ConfirmStackModal.tsx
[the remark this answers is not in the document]
berteotti: Fixed in c9a830dae34db6f75e49a63d9b2c0a972db77ef6

packages/app/components/stackbox/ConfirmStackModal.tsx
```diff
@@ -13,30 +13,76 @@ import {
...
+  startTime: Date;
+  endTime: Date;
+}
+
+const frequencyTitles = {
+  [FREQUENCY_OPTIONS.hour]: "hour",
+  [FREQUENCY_OPTIONS.day]: "day",
+  [FREQUENCY_OPTIONS.week]: "week",
+  [FREQUENCY_OPTIONS.month]: "month",
+};
```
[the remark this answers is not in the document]
berteotti: Fixed in c9a830dae34db6f75e49a63d9b2c0a972db77ef6

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -60,39 +60,67 @@ const balanceOptions = [
...
-  const openConfirmStack = () => openModal(ModalId.CONFIRM_STACK);
-  const openTokenPicker = (isTokenFrom = true) => {
-    setIsPickingTokenFrom(isTokenFrom);
-    openModal(ModalId.TOKEN_PICKER);
-  };
-  const selectToken = isPickingTokenFrom ? setTokenFrom : setTokenTo;
+  const [showTokenAmountError, setShowTokenAmountError] = useState(false);
+  const [showDateTimeError, setShowDateTimeError] = useState(false);
+  const [showFromTokenError, setShowFromTokenError] = useState(false);
+  const [showToTokenError, setShowToTokenError] = useState(false);
```
ElRodrigote, line 90: WDYT about dropping these `useState` variables and storing the error styles in a variable for readability purposes? Feels easier than toggling on and off checking other state variables.

Maybe I'm missing something tho.

berteotti: What do you mean with error styles in a variable? I can only see `useReducer` simplifying this.

berteotti: Oh, you mean having a style directly defined in the variable instead of a bool value. That also can work. But can't use a `useReducer` on it. Don't know which path is better.

ElRodrigote: Which reducer are you talking about? D: 
I can't find it.

On my suggestion about the styles into a variable, it's because I thought you created these `useState` variables for mostly conditional rendering and/or conditional styling (to keep conditional styles names meaningfully).

What I see is:
`showTokenAmountError` -> conditional styling
`showDateTimeError` -> conditional rendering
`showFromTokenError` & `showToTokenError` -> conditional styling + animation event trigger

The only place where these values go `true` is when the `Stack now` button gets clicked but all the 5 values (the 2 dates and the token values) are either coming from a hook or stateful variables.

So my doubt is around why do we need these 4 error stateful variables if we could simply use their respective dependency values.

But maybe I'm missing something like the reducer you're talking about but I can't find it D:

berteotti: I'm talking about `useReducer` as a way to improve this code :)

packages/app/components/stackbox/ConfirmStackModal.tsx
Diogomartf: Makes sense.

Close when it's only used there, but move it to a model when used in more places.

packages/app/components/stackbox/Stackbox.tsx
berteotti: I think we can had a `useReducer` in the future. It will make the component cleaner. For now went with you suggestion! Tell me what you think.

packages/app/models/stack/types.ts
Diogomartf: Thanks for sharing your thoughts @ElRodrigote.  

I agree with you guys.

Diogomartf:
I didn't like the last update, this commit [cbb858a](https://github.com/SwaprHQ/stackly-ui/pull/42/commits/cbb858aa4cafca82c5e347be5c34a72b25e84cc4),  of storing the classNames on the state.

And if you want to do that you also have to change the state variables to reflect that, `showTokenAmountError` now is not a boolean, so it would have to change to `showTokenAmountErrorClassname` or something like that to reflect that.

BUT I think that change it's not good, I think it's clear to have the classNames on the element it gives styles and use state for controlling the conditional with booleans.

For example, here, I think on the "green" it's less clear what's going on, while on the "red" I know we are changing state flow.
image

Also here is not a place to have the styles, if I'm looking to change some error style, I'll go look for the element and not the state flow declaration.

It's mixing things. I would revert this commit.

packages/app/app/ui/page.tsx
```diff
@@ -26,6 +26,7 @@ import {
   Toast,
   Severity,
 } from "@/ui";
+import { FREQUENCY_OPTIONS } from "../../models/stack";
```
Diogomartf, line 29: can we import like this? 

```suggestion
import { FREQUENCY_OPTIONS } from "@/models/stack";
```

berteotti:
> I didn't like the last update, this commit [cbb858a](https://github.com/SwaprHQ/stackly-ui/pull/42/commits/cbb858aa4cafca82c5e347be5c34a72b25e84cc4), of storing the classNames on the state.
> 
> And if you want to do that you also have to change the state variables to reflect that, `showTokenAmountError` now is not a boolean, so it would have to change to `showTokenAmountErrorClassname` or something like that to reflect that.
> 
> BUT I think that change it's not good, I think it's clear to have the classNames on the element it gives styles and use state for controlling the conditional with booleans.
> 
> For example, here, I think on the "green" it's less clear what's going on, while on the "red" I know we are changing state flow. image
> 
> Also here is not a place to have the styles, if I'm looking to change some error style, I'll go look for the element and not the state flow declaration.
> 
> It's mixing things. I would revert this commit.
> 
> I think what @ElRodrigote suggested was not this, but reduce the state variables and use the values directly like using the already defined `!fromToken` instead of creating a new state `showFromTokenError`. I think it makes sense but we might be missing something.

I don't like it either, it mixes things and we are applying styles on hooks. We should only apply styles on classNames IMO.

Ok, so joining 1 error state + variable states. I think we can't do that because there are animations with different durations. So "Select token" animation ends before input animation. If we have a global error state all animation will end at the same time. Plus, date time error will disappear after a few seconds.

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -21,32 +21,33 @@ import {
 } from "@/components";
 import { TokenFromTokenlist } from "@/models/token";
 import { useAccount, useBalance, useNetwork } from "wagmi";
-import { formatUnits } from "viem";
+import { formatUnits, parseUnits } from "viem";
 import { ModalId, useModalContext } from "@/contexts";
+import { FREQUENCY_OPTIONS } from "@/models/stack";
+import { BigNumber } from "ethers";
```
Diogomartf, line 27: can't we use native `BigInt` ?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt

viem docs: https://viem.sh/docs/ethers-migration.html#bignumber-utilities

Diogomartf: It will be handier if we leave `@ethers` lib

Diogomartf:
IMO this experience got worse, too many things shaking.

7ac1caba2c6092fc197fd22c67f59001

What I think will improve it:
- if there is no token selected or just one token, only shake the "select token" button(s)

Diogomartf:
When no connected wallet this button should be the connect button or something like that, what do you think?

image

image

_This might be out of scope of this PR but should be implemented on "creating stack" flow of PRs._

packages/app/components/stackbox/Stackbox.tsx
berteotti: We actually can, thanks!

berteotti:
> IMO this experience got worse, too many things shaking.
> 
> 7ac1caba2c6092fc197fd22c67f59001
> 
> What I think will improve it:
> 
> * if there is no token selected or just one token, only shake the "select token" button(s) (not shake the place holder in this case).
> 
> The rest of the flow seems alright.

Although I think this solution hides errors, I think it's a cleaner experience. Will update it.

Diogomartf:
I think it's looking great!

The animation 👌 

PS: The modal is not opening, probably related with the merged PR that adds support for multiple modals.

Diogomartf (approved):

packages/app/components/stackbox/Stackbox.tsx
ElRodrigote: Ohhh I get now what you said. Yeah, I'm ok with that.

I saw the discussion on the changes, it's my bad for poorly explaining what I had in mind, next time I'll bring a block of code as a suggestion with the explanation, sorry for the extra work :c

ElRodrigote (approved):

PR #43, opened 2023-08-10
fix: condition got error when cowOrders was undefined

Diogomartf:
**Description**
Fix error when opening a modal with cowOrders undefined

**Summary**
- add it to calculateStackAveragePrice and totalStackOrdersDone

**Error**
image

### DIFF
--- a/packages/app/models/stack-order/stack-order.ts
+++ b/packages/app/models/stack-order/stack-order.ts
@@ -1,14 +1,17 @@
 import { StackOrder } from "@/models/stack-order";
 import { convertedAmount } from "@/utils/numbers";
 
-export const totalStackOrdersDone = (order: StackOrder) =>
-  order.cowOrders.length;
+export const totalStackOrdersDone = (order: StackOrder) => {
+  if (!order?.cowOrders?.length) return 0;
+
+  return order.cowOrders.length;
+};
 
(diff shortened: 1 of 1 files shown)


packages/app/models/stack-order/stack-order.ts
```diff
@@ -1,14 +1,20 @@
...
+
+  return order.cowOrders.length;
+};
 
 export const calculateStackAveragePrice = (order: StackOrder) => {
   let totalExecutedBuyAmount = 0;
   let totalExecutedSellAmount = 0;
 
-  if (order.cowOrders.length === 0) return 0;
+  if (noCowOrders(order)) return 0;
```
berteotti, line 17: This looks simpler to me

```suggestion
  if (!order?.cowOrders?.length) return 0;
```

ElRodrigote (dismissed):
I agree with Leo's comment, but I'll leave my approve as it is a small enough fix either way

packages/app/models/stack-order/stack-order.ts
Diogomartf: I'll test it, it it works, I'm down with this

Diogomartf: Fixed!

berteotti (approved):

PR #44, opened 2023-08-10
feat: Add support for multiple modals open

Diogomartf:
**Description**
- Update `ModalContext` to support multiple modals open
- Update components to use new api

_Removed Toast for now from `ModalsContext` based on @ElRodrigote it shouldn't be here, we can leave it open for now as it's not added to the flow yet._

### DIFF
--- a/packages/app/components/stackbox/Stackbox.tsx
+++ b/packages/app/components/stackbox/Stackbox.tsx
@@ -63,7 +63,7 @@ export const Stackbox = () => {
   const [isPickingTokenFrom, setIsPickingTokenFrom] = useState<boolean>(false);
   const [tokenFrom, setTokenFrom] = useState<TokenFromTokenlist>();
   const [tokenTo, setTokenTo] = useState<TokenFromTokenlist>();
-  const { closeModal, openModalId, openModal } = useModalContext();
+  const { closeModal, isModalOpen, openModal } = useModalContext();
 
   const { chain } = useNetwork();
   const { address } = useAccount();
@@ -126,7 +126,7 @@ export const Stackbox = () => {
           />
           <Icon
             name="arrow-left"
-            className="flex items-center justify-center p-2 w-16 h-9 bg-surface-50 rounded-2xl rotate-180"
+            className="flex items-center justify-center w-16 p-2 rotate-180 h-9 bg-surface-50 rounded-2xl"
           />
           <SelectTokenButton
             label="To receive"
@@ -139,7 +139,7 @@ export const Stackbox = () => {
             type="number"
             pattern="[0-9]*"
             placeholder="0.0"
-            className="w-full py-3 text-4xl text-em-med font-semibold outline-none"
+            className="w-full py-3 text-4xl font-semibold outline-none text-em-med"
             value={tokenAmount}
             onKeyDown={(evt) =>
               ["e", "E", "+", "-"].includes(evt.key) && evt.preventDefault()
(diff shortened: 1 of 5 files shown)


berteotti (approved):
Awesome, lgtm!

packages/app/contexts/ModalContext.tsx
```diff
@@ -1,42 +1,88 @@
...
+  closeModal: (id: ModalId) => {},
   openModal: (id: ModalId) => {},
+  isModalOpen: (id: ModalId) => false,
 });
 
 export interface ModalContextProviderProps {
   children: ReactNode;
 }
 
+type Action = { id: ModalId; type: "add" | "remove" };
```
ElRodrigote, line 41:
```suggestion
type Action = { id: ModalId; type: ActionType };
```

packages/app/contexts/ModalContext.tsx
```diff
@@ -1,42 +1,88 @@
...
     () => ({
-      closeModal: () => setOpenModalId(null),
-      openModal: (id: ModalId) => setOpenModalId(id),
-      openModalId,
+      closeModal,
+      openModal,
+      isModalOpen,
     }),
-    [openModalId]
+    [isModalOpen]
```
ElRodrigote, line 85: Please try this out if it works, it may improve the perf in this context. Each time we render something we re-define this `isModalOpen` variable and with that re-definition it will re-trigger the `useMemo` as it's a dependency there (kinda beating the purpose of memorizing the context values)
```tsx
  const modalContext = useMemo(() => {
    const isModalOpen = (modalId: ModalId) => openModals.includes(modalId);

    return {
      closeModal,
      openModal,
      isModalOpen,
    };
  }, [openModals]);
```

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -237,25 +237,25 @@ export const Stackbox = () => {
...
+        isOpen={isModalOpen(ModalId.TOKEN_PICKER)}
         onTokenSelect={selectToken}
       />
       <ConfirmStackModal
-        isOpen={openModalId === ModalId.CONFIRM_STACK}
-        closeAction={closeModal}
+        isOpen={isModalOpen(ModalId.CONFIRM_STACK)}
+        closeAction={() => closeModal(ModalId.CONFIRM_STACK)}
       />
       <Toast
```
ElRodrigote, line 249: Thinking out loud and not sure we should even have this toast here now that I think about it. Unless we just want to lay some sort of groundwork when we decide to add it to the flow.

berteotti: Had the same concern, but now that we have it we can leave it as it is. But I prefer in the future if we don't leave unused/mock components throughout the layout.

Diogomartf: makes sense, I think I can remove it.

packages/app/contexts/ModalContext.tsx
Diogomartf: yeah true, this one missing piece that I wanted your suggestion but forgot to ask. 

Thanks, that's the change we needed.

packages/app/contexts/ModalContext.tsx
Diogomartf: good catch

berteotti (approved):
Nice tip from @ElRodrigote . Looks even better now!

PR #45, opened 2023-08-11
fix: Progress bar should be based on cow orders and not on order slots

Diogomartf:
**Description**
Progress bar should be based on cow orders and not on order slots, this was causing a bug of showing progress without orders.

**Screenshot**
image


Diogomartf:
When no connected wallet this button should be the connect button or something like that, what do you think?

image

image

_This is probably out of scope of this PR but should be implemented on "creating stack" flow of PRs._

berteotti (approved):

PR #47, opened 2023-08-14
feat: show when a stack is finished but has funds

Diogomartf:
**Description**
There are stacks that not complete all the orders and the stack has remaining funds this creates the need to withdraw the funds manually. 

This PR aims to make clear when the stack is on that state.

**images**
image
image


packages/app/components/stack-modal/StackModal.tsx
```diff
@@ -74,17 +76,36 @@ export const StackModal = ({
...
-  const remainingFundsDescription =
-    remainingFunds > 0 &&
-    `remaining funds, ${remainingFunds} ${stackOrder.sellToken.symbol}`;
+    return confirmCancelContent.notComplete;
+  };
+
+  const remainingFundsText = () =>
+    `The ${stackRemainingFundsWithTokenText(
+      stackOrder
+    )} will be sent to your wallet.`;
```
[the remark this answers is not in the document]
Diogomartf: what's the suggestion, using it directly?

packages/app/models/stack-order/stack-order.ts
```diff
@@ -43,3 +44,27 @@ export const totalStacked = (order: StackOrder) =>
...
+export const stackIsComplete = (stackOrder: StackOrder) =>
+  allOrderSlotsDone(stackOrder) && !stackHasRemainingFunds(stackOrder);
+
+export const stackIsFinishedWithFunds = (stackOrder: StackOrder) =>
+  allOrderSlotsDone(stackOrder) &&
+  stackHasRemainingFunds(stackOrder) &&
+  !stackOrder.cancelledAt;
+
+export const stackRemainingFundsWithTokenText = (stackOrder: StackOrder) =>
+  `${stackRemainingFunds(stackOrder)} ${stackOrder.sellToken.symbol}`;
```
berteotti, line 70: If we declare this inside the component, there is no need to be a function. 

```suggestion
export const stackRemainingFundsWithTokenText = `${stackRemainingFunds(stackOrder)} ${stackOrder.sellToken.symbol}`;
```

This will work inside the component, we just need to call `stackRemainingFundsWithTokenText` and not `stackRemainingFundsWithTokenText(stackOrder)`

berteotti: I recommend this because I don't see anywhere else we are calling this function.

packages/app/components/stack-modal/StackModal.tsx
```diff
@@ -174,14 +204,18 @@ export const StackModal = ({
...
         <ModalFooter>
-          {!stackOrder.cancelledAt && !stackIsComplete && (
+          {!stackOrder.cancelledAt && !stackIsComplete(stackOrder) && (
             <Button
               size="sm"
-              action="secondary"
+              action={
+                getConfirmCancelContent().button.action as
+                  | "primary"
+                  | "secondary"
```
berteotti, line 213: It's better to add types to `confirmCancelContent` instead of casting them here

packages/app/components/stack-modal/StackModal.tsx
```diff
@@ -74,17 +76,36 @@ export const StackModal = ({
...
+    },
+    finishedWithFunds: {
+      title: "Proceed with cancelation to withdraw funds",
+      description: remainingFundsText(),
+      button: {
+        action: "primary",
+        text: `Withdraw ${stackRemainingFundsWithTokenText(stackOrder)}`,
+      },
+    },
+  };
```
berteotti, line 108: This can be 2 separate variables. I see no reason to have 1 variable with 2 distinct keys

packages/app/models/stack-order/stack-order.ts
Diogomartf: makes sense, I'll fix it.

packages/app/components/stack-modal/StackModal.tsx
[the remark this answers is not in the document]
Diogomartf: Yes, I'll remove the function, it's just a variable.

packages/app/components/stack-modal/StackModal.tsx
Diogomartf: It can

Diogomartf: fixed.

packages/app/components/stack-modal/StackModal.tsx
```diff
@@ -74,17 +85,32 @@ export const StackModal = ({
...
 
-  const remainingFundsDescription =
-    remainingFunds > 0 &&
-    `remaining funds, ${remainingFunds} ${stackOrder.sellToken.symbol}`;
+  const remainingFundsText = `The ${stackRemainingFundsWithTokenText} will be sent to your wallet.`;
 
-  const stackIsComplete =
-    totalOrderSlotsDone(stackOrder) === orderSlots.length &&
-    remainingFunds === 0;
+  const getConfirmCancelContent = (): Content => {
```
ElRodrigote, line 94: **Minor, non-blocking**:
`description` property can be abstracted into its own variable as it's the same value

ElRodrigote (approved):

packages/app/components/stack-modal/StackModal.tsx
Diogomartf: It could but don't think it will be much better. 

I like the fact if we want to change anything about the dialog it's all here, otherwise we'll be changing the title on this object and for the description we'll have to go to the component. 

I'll let it be here for now.

packages/app/components/stack-modal/StackModal.tsx
```diff
@@ -50,6 +52,17 @@ interface StackModalProps extends ModalBaseProps {
...
+  button: {
+    action: "primary" | "secondary";
+    text: string;
+  };
+};
+
+export const txEplorerLink = (tx: string) => `https://gnosisscan.io/tx/${tx}`;
+
 export const addressExplorerLink = (address: string) =>
   `https://gnosisscan.io/address/${address}#tokentxns`;
```
berteotti, line 67: If we are using this elsewhere we might move this to @/model

berteotti (approved):
Looking good, just made a suggestion

packages/app/components/stack-modal/StackModal.tsx
Diogomartf: removed this method, as we have one on `utils/transaction` file

PR #54, opened 2023-08-21
[STK-105][FEAT] - Display user's token balances

ElRodrigote:
### Description

* Updated Modal logic to opt-out the background opacity
* Created the Toast component
* Updated UI page where we display Stackly components

### Preview
image
image

### How to test the changes
1) Pull this branch
2) Run the project locally
3) Go to the Stackly homepage, connect your wallet, and then click any of the two "Select token" buttons

### DIFF
--- a/packages/app/components/token-picker/TokenPicker.tsx
+++ b/packages/app/components/token-picker/TokenPicker.tsx
@@ -1,6 +1,7 @@
 "use client";
 
 import { ChangeEvent, RefObject, forwardRef, useEffect, useState } from "react";
+import { useAccount, useBalance, useNetwork } from "wagmi";
 
 import {
   BodyText,
@@ -17,6 +18,7 @@ import { TokenFromTokenlist } from "@/models/token/types";
 import { TokenIcon } from "@/components";
 import { TOKEN_PICKER_COMMON_TOKENS } from "./constants";
 import { useTokenListContext } from "@/contexts";
+import { formatTokenValue } from "@/utils/token";
 
 const HALF_SECOND = 500;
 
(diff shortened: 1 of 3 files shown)


packages/app/components/token-picker/TokenPicker.tsx
```diff
@@ -152,26 +154,44 @@ const CommonTokens = ({ onTokenSelect }: CommonTokensProps) => (
...
+  return (
+    <div
+      className="flex justify-between w-full py-2 cursor-pointer hover:bg-surface-50"
+      key={token.address}
+      onClick={() => onTokenSelect(token)}
+    >
+      <div className="flex items-center justify-between w-full">
+        <div className="flex items-center space-x-3">
+          <TokenIcon token={token} size="md" />
+          <div className="flex flex-col">
```
Diogomartf, line 179: It works the same if we remove the styles

```suggestion
          <div>
```

packages/app/components/token-picker/TokenPicker.tsx
```diff
@@ -152,26 +154,44 @@ const CommonTokens = ({ onTokenSelect }: CommonTokensProps) => (
...
-            {token.name}
-          </BodyText>
+const TokenListRow = ({ onTokenSelect, token }: TokenListRowProps) => {
+  const { chain } = useNetwork();
+  const { address } = useAccount();
+  const { data: balance } = useBalance({
+    address,
+    token: token?.address as `0x$string`,
+    chainId: chain?.id,
+  });
```
berteotti, line 164: I think we can push this for now, it was slow in the first run, but then it used the fetched values.

packages/app/utils/token.ts
```diff
@@ -1,8 +1,31 @@
...
+  const nonRelevantZeroes = getZeroDecimalsCount(value);
+  const relevantZeroDecimals = new Array(decimals - 1).fill(0).join("");
+
+  const formattedValue =
+    nonRelevantZeroes >= decimals
+      ? `< 0.${relevantZeroDecimals}1`
+      : Number(value).toFixed(decimals);
+
+  return formattedValue;
 };
```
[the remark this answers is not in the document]
ElRodrigote: We can dumb down the logic as long as we always use 4 decimals, but the idea for `getZeroDecimalsCount` is to get the number of zeroes before the first relevant magnitude. It's used mainly to format the dust values.

Now that I check it, the function is wrong, because that format should apply only for near zero values and not for any value.

WDYT? For now, it's working with any given amount of decimals (so if we eventually need to format tokens with 6 decimal places we can do it too.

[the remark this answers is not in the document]
ElRodrigote: The new change I just pushed would yield
`1.000000000000000001` -> `1.0000`
`0.000000000000000001` -> `< 0.0001`

packages/app/utils/token.ts
```diff
@@ -1,8 +1,31 @@
...
-  return custom ? value.toFixed(custom) : value.toFixed(4);
+  const nonRelevantZeroes = getZeroDecimalsCount(value);
+  const relevantZeroDecimals = new Array(decimals - 1).fill(0).join("");
+
+  const formattedValue =
+    nonRelevantZeroes >= decimals
+      ? `< 0.${relevantZeroDecimals}1`
+      : Number(value).toFixed(decimals);
+
+  return formattedValue;
```
berteotti:
```suggestion
  if (value === 0) return value;

  const fixedValue = (
    Math.floor(parseFloat(value as string) * 100) / 100
  ).toFixed(decimals);
  const parsedFixedValue = parseFloat(fixedValue);
  const relevantZeroDecimals = new Array(decimals - 1).fill(0).join("");

  const formattedValue =
    parsedFixedValue === 0
      ? `< 0.${relevantZeroDecimals}1`
      : Number(value).toFixed(decimals);

  return formattedValue;
```
I was suggesting something like this. We don't need to go through the array finding indexes. We can just cast it twice to know if it's lower than 0.0001.

ElRodrigote: I've been doing a few runs with test values but at least on my end, it seems to not be working the same way. All these samples ran with 4 decimals (default for formatting for now) and different magnitudes:
* 0.001 (higher than 0.0001)
image
* 0.0001 (equals to 0.0001)
image
* 0.00001 (smaller than 0.0001, this result is ok)
image

ElRodrigote: After an offline talk with Leo, we decided for a less tricky and simpler way to achieve this formatting.

packages/app/utils/token.ts
```diff
@@ -1,8 +1,18 @@
 export const addressShortner = (address: string) =>
   address.slice(0, 4) + "..." + address.slice(-4);
 
-export const formatTokenValue = (value: number, custom?: number) => {
+export const formatTokenValue = (
+  value: number | string,
+  decimals: number = 4
+) => {
   if (value === 0) return value.toFixed(2);
```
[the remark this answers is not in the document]
ElRodrigote: Oh, this is legacy code, I didn't touch it because it may be used in another place (as diogo did it).

What I'm doing is to check [inside the TokenRow](https://github.com/SwaprHQ/stackly-ui/pull/54/files#diff-44dd933316d2da92fe4ba3e95088dd98d164ade168363074c5c5a436656a469fR189) if I should display a `0` or a formatted value

PR #60, opened 2023-08-22
fix: outline stroke on radio buttons should be lighter

Diogomartf:
image

comments: https://web.ruttl.com/share/gWnw47S1jJZvzZcB4eez


ElRodrigote (dismissed):

berteotti (changes requested):
I think this should be applied to every button, not only radio buttons

Diogomartf:
> I think this should be applied to every button, not only radio buttons

I think this is a bit out of scope of this PR, but as is it is just the ring, we can make them lighter on the buttons as well.

>

berteotti (approved):
LGTM. IMO Radio Buttons and Buttons shouldn't be different, except for the border radius. In my view Akash comments was made for every button.

PR #62, opened 2023-08-24
feat: set token amount to balance when clicking the balance

Diogomartf:
Akash gave a good suggestion about setting the balance to max when clicking the balance.

This PR implements that.

https://github.com/SwaprHQ/stackly-ui/assets/5664434/cdbd05fc-a82f-4ca1-a768-e9c9911b3903


packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -222,27 +232,24 @@ export const Stackbox = () => {
...
                 <TokenIcon token={fromToken} size="2xs" />
-                <BodyText className="text-em-high">
-                  <span className="text-em-low">Balance:</span>{" "}
-                  {formattedBalance(balance)}
-                </BodyText>
+                <div className="flex items-center space-x-1">
+                  <BodyText className="flex items-center text-em-low">
+                    Balance:
+                  </BodyText>
+                  <div onClick={() => setTokenAmountBasedOnBalance(1)}>
```
berteotti, line 247: Maybe we should have a name for this 1. `Balance.MAX` for example

Diogomartf: let's name them all.

berteotti: Yes, if we name 1 we name them all :)

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -52,10 +52,16 @@ const startDateTimeTimestamp = new Date().setMinutes(
   new Date().getMinutes() + 30
 );
 
+const BalanceDivider = {
+  max: 1,
+  half: 2,
+  quarter: 4,
```
berteotti, line 58: Can we make it uppercase to show that's a constant? Was thinking about using an enum but this works fine too.

Diogomartf: I can.

Diogomartf: fixed!

berteotti (approved):
Good job Diogo!

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -214,35 +230,36 @@ export const Stackbox = () => {
...
-                </BodyText>
+                <div className="flex items-center space-x-1">
+                  <BodyText className="flex items-center text-em-low">
+                    Balance:
+                  </BodyText>
+                  <div
+                    onClick={() =>
+                      setTokenAmountBasedOnBalance(BalanceDivider.MAX)
+                    }
+                  >
```
berteotti, line 257: We shouldn't use divs with onClick, this should be a button.

berteotti (changes requested):
Forgot to check the button

packages/app/components/stackbox/Stackbox.tsx
berteotti: Also, we can declare the button inside the `BodyText` and just apply a `className="cursor-pointer hover:underline text-em-high"` to it.

berteotti (approved):

PR #65, opened 2023-08-24
update: move page padding to layout

Diogomartf:
fixes:
- move padding to layout 
- icon was overlaping menu on mobile

### DIFF
--- a/packages/app/components/navbar/MobileMenu.tsx
+++ b/packages/app/components/navbar/MobileMenu.tsx
@@ -11,7 +11,7 @@ export default function MobileMenu() {
   const toggle = () => setIsOpen(!isOpen);
 
   return (
-    <div className="flex items-center justify-end w-full gap-4 md:hidden">
+    <div className="z-10 flex items-center justify-end w-full gap-4 md:hidden">
       <SelectNetwork />
       <ConnectButton />
       <Button
--- a/packages/app/app/layout.tsx
+++ b/packages/app/app/layout.tsx
@@ -23,10 +23,10 @@ export const metadata: Metadata = {
 export default function RootLayout({ children }: PropsWithChildren) {
   return (
     <html lang="en">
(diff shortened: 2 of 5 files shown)


berteotti:
image

Navbar has padding in mobile. I think the solution is to give padding to Navbar and around the children inside layout. Navbar padding will keep the bottom border, the padding around the children will fix the missing padding on mobile.

Diogomartf:
> image
> 
> Navbar has padding in mobile. I think the solution is to give padding to Navbar and around the children inside layout. Navbar padding will keep the bottom border, the padding around the children will fix the missing padding on mobile.

Yeah, added padding to children and navbar as they not share the same pattern.

So will always have a padding x of 16px on children, and Navbar takes care of its padding.

packages/app/components/navbar/MobileMenu.tsx
```diff
@@ -11,7 +11,7 @@ export default function MobileMenu() {
   const toggle = () => setIsOpen(!isOpen);
 
   return (
-    <div className="flex items-center justify-end w-full gap-4 md:hidden">
+    <div className="z-10 flex items-center justify-end w-full gap-4 md:hidden">
```
berteotti, line 14: Why this z index?
This is happening with this change. What should we do here? We can increase the overlay z index if needed

Screenshot 2023-08-25 at 14 21 48

Diogomartf: image

The idea was to fix the arrow from overlapping the menu. Will try to find a solution.

Diogomartf: fixed [c5423a6](https://github.com/SwaprHQ/stackly-ui/pull/65/commits/c5423a6ecb804f003b7cb41d7f36c5da3c0acafb)

berteotti (approved):

ElRodrigote (approved):

PR #73, opened 2023-08-30
feat: create mint page

berteotti:
Create mint page without contract logic. Mint NFT section has different spacing from figma.

Screenshot 2023-08-30 at 14 45 37

### DIFF
--- a/packages/app/app/mint/page.tsx
+++ b/packages/app/app/mint/page.tsx
@@ -1,0 +1,61 @@
+"use client";
+
+import { useState } from "react";
+import Image from "next/image";
+import { StacklyBetaNFTImg } from "@/public/assets";
+import { BodyText, Button, ButtonLink, HeadingText } from "@/ui";
+
+export default function Page() {
+  const [hasMinted, setHasMinted] = useState(false);
+
+  return (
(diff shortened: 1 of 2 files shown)


Diogomartf:
NFT image is too big, should have a max width about 420px, something like: `max-w-md` or `max-w-sm`.

image

Diogomartf:
Mobile view is not ideal, would be better to have the mint button at the bottom, but that would requires us to use a grid and might be a lot more work for the usage.

image

Diogomartf (approved):

berteotti:
> Mobile view is not ideal, would be better to have the mint button at the bottom, but that would requires us to use a grid and might be a lot more work for the usage.
> 
> image

We can put the button on the bottom. But what about UX? Scrolling to mint doesn't seem a good experience to me. I just want to mint so I can Stack.

berteotti:
Should we change the text to `Mint (FREE)` ?

Diogomartf:
> > Mobile view is not ideal, would be better to have the mint button at the bottom, but that would requires us to use a grid and might be a lot more work for the usage.
> > image
> 
> We can put the button on the bottom. But what about UX? Scrolling to mint doesn't seem a good experience to me. I just want to mint so I can Stack.

okay, I'll buy that.

PR #75, opened 2023-08-31
feat: mint Beta NFT logic

berteotti:
This PR adds NFT logic to mint page and index page (to show the modal if the user has no NFT).


packages/app/app/mint/page.tsx
```diff
@@ -1,61 +1,170 @@
...
+        nftWhitelistMaxSupply(nftWhitelist),
+      ])
+        .then(([newMintedAmount, newMaxSupply]) => {
+          if (newMintedAmount.status === "fulfilled")
+            setMintedAmount(newMintedAmount.value.toString());
+          if (newMaxSupply.status === "fulfilled")
+            setMaxSupply(newMaxSupply.value.toString());
+        })
+        .catch((e) =>
+          console.error("Error while fetching NFT contract info: ", e)
```
Diogomartf, line 53: should we `setError` here to give feedback to user?

packages/app/app/mint/page.tsx
```diff
@@ -1,61 +1,170 @@
...
+    };
+
+    fetchData();
+  }, [nftWhitelist]);
+
+  useEffect(() => {
+    const fetchAddressBalance = async () => {
+      if (!address) {
+        return;
+      }
```
Diogomartf, line 64:
```suggestion
      if (!address) return;
```

packages/app/app/mint/page.tsx
```diff
@@ -1,61 +1,170 @@
...
+    if (isDisconnected) {
+      setError(undefined);
+      setIsNFTHolder(false);
+    }
+  }, [isDisconnected]);
+
+  const mint = async () => {
+    if (!signer || !address || !chain) {
+      return;
+    }
```
Diogomartf, line 89:
```suggestion
    if (!signer || !address || !chain) return;
```

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -179,9 +186,26 @@ export const Stackbox = () => {
...
+  }, [closeModal, isConnected, isModalOpen]);
+
+  useEffect(() => {
+    const getNFTHolder = async () => {
+      if (address && chain && signer) {
+        const nftWhitelist = getWhitelist(
+          getNftWhitelistAddress(chain.id),
+          signer
+        );
+        const amount = await nftWhitelistBalanceOf(nftWhitelist, address);
```
Diogomartf, line 200:
```suggestion
        const nftAmount = await nftWhitelistBalanceOf(nftWhitelist, address);
```

packages/app/contexts/ModalContext.tsx
```diff
@@ -9,6 +9,7 @@ import {
 } from "react";
 
 export enum ModalId {
+  WAITING_TRANSACTION = "waiting_transaction",
```
Diogomartf, line 12: Knowing here that's the NFT related transaction is good.

```suggestion
  NFT_WAITING_TRANSACTION = "nft_waiting_transaction",
```

packages/sdk/src/vaults/whitelist.ts
Diogomartf, line 1: this adds complexity to our SDK (if one day others want to use it) but I guess it's a necessary evil at this point.

packages/app/app/mint/page.tsx
berteotti: No, this error isn't related to minting the NFT. User can still mint the NFT if this throws an error.

packages/app/contexts/ModalContext.tsx
berteotti: I think we can have a default waiting transaction state. What do you think?

packages/sdk/src/vaults/whitelist.ts
berteotti: We need those to interact with the contracts.

packages/app/contexts/ModalContext.tsx
Diogomartf: We currently have: `STACK_APPROVE_PROCESSING`, `CANCEL_STACK_PROCESSING`, `STACK_CREATION_PROCESSING`

to be consistent would be:`NFT_MINT_PROCESSING`

I don't know if using the `TRANSACTION_PROCESSING` or `WAITING_TRANSACTION` in all of them is better. 
But I'll let you decide on that.

packages/sdk/src/vaults/whitelist.ts
Diogomartf: I know, but they add a bit of noise to `SDK`. I just want that we are aware of that.

packages/app/contexts/ModalContext.tsx
berteotti: Will change to something more specific.

Diogomartf (approved):

PR #76, opened 2023-08-31
fix: padding on StackModal, default color and more

Diogomartf:
**summary**
- add default color to body
- add lighter border color to "tertiary" button
- check modal padding x (24px)
- stack modal header padding y
- consistent padding inside Stack modal

**images**
image
image

### DIFF
--- a/packages/app/components/stack-modal/StackModal.tsx
+++ b/packages/app/components/stack-modal/StackModal.tsx
@@ -165,7 +165,7 @@ export const StackModal = ({
             />
           </div>
         </ModalHeader>
-        <ModalContent className="px-0 md:px-0">
+        <ModalContent className="px-0 space-y-4 md:px-0">
           <div className="grid grid-cols-2 gap-5 px-4 md:px-6 gap-x-8 md:grid-cols-4">
             <StackDetail title="Starts on">
               {formatTimestampToDateWithTime(firstSlot)}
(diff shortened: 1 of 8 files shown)


packages/app/components/stack-modal/StackModal.tsx
```diff
@@ -205,18 +205,19 @@ export const StackModal = ({
...
               onClick={() => openModal(ModalId.CANCEL_STACK_CONFIRM)}
               width="full"
             >
               {getConfirmCancelContent().button.text}
             </Button>
-          )}
-        </ModalFooter>
+          </ModalFooter>
+        ) : (
+          <div className="pb-6"></div>
```
berteotti, line 219: Shouldn't we apply a style to modal instead of adding an empty div?

ElRodrigote: I'd rather keep the short-circuit logic and add a conditional styling (`pb-6`) to the content rather than having an empty div TBH.

Something like abstracting `!stackOrder.cancelledAt && !stackIsComplete(stackOrder)` into a variable and use the same variable for the short-circuit AND the conditional styling for the padding

ElRodrigote (dismissed):

packages/app/components/stack-modal/StackModal.tsx
Diogomartf: Did extract it into a variable, `stackNotCancelledAndNotComplete`

berteotti (approved):

PR #83, opened 2023-09-04
[STK-141][FEAT] - Add start datetime logic

ElRodrigote:
## Fixes: 
* [STK-141](https://linear.app/swaprhq/issue/STK-141/add-start-datetime-logic)
* [STK-137](https://linear.app/swaprhq/issue/STK-137/show-error-when-creating-stack-in-the-past)

### Description

* Check if the start date is 10m or less from now, if so, display an error
* Highlight date fields that are erroring
* Allow 1 date field error at the time

### Preview
image
image

### How to test the changes
1) Pull this branch
2) Run the project locally
3) Go to the Stackly homepage, connect your wallet
4) The current date time should be displayed in the "start time" input.
4.1) If the user selects a date that is less than 10 minutes from the current time, an error should appear
4.2) If the user has both errors for start and end date, then only 1 error at the time should be displayed

### DIFF
--- a/packages/app/components/stackbox/Stackbox.tsx
+++ b/packages/app/components/stackbox/Stackbox.tsx
@@ -46,6 +46,8 @@ interface SelectTokenButtonProps {
 }
 
 const START_TIME_MINUTES_OFFSET = 10;
+const getDateNowPlus10Mins = () =>
+  new Date().setMinutes(new Date().getMinutes() + START_TIME_MINUTES_OFFSET);
 
 const frequencyOptions = [
   { option: FREQUENCY_OPTIONS.hour, name: "Hour" },
@@ -104,7 +106,7 @@ export const Stackbox = () => {
 
   const [showTokenAmountError, setShowTokenAmountError] = useState(false);
   const [showPastEndDateError, setShowPastEndDateError] = useState(false);
-  const [showPastStartDateError, setShowPastStartDateError] = useState(false);
+  const [isPastStartDate, setIsPastStartDate] = useState(false);
   const [showFromTokenError, setShowFromTokenError] = useState(false);
   const [showToTokenError, setShowToTokenError] = useState(false);
   const [showInsufficentBalanceError, setShowInsufficentBalanceError] =
@@ -123,15 +125,11 @@ export const Stackbox = () => {
     const startDate = startDateTime.getTime();
     const endDate = endDateTime.getTime();
     const isEndTimeBeforeStartTime = endDate <= startDate;
-    const isStartTimeBefore10minsFromNow =
-      startDate <=
-      new Date().setMinutes(
-        new Date().getMinutes() + START_TIME_MINUTES_OFFSET
-      );
+    const isStartTimeInThePast = startDate <= Date.now();
     const isTokenAmountZero = tokenAmount === "0";
 
     setShowPastEndDateError(isEndTimeBeforeStartTime);
-    setShowPastStartDateError(isStartTimeBefore10minsFromNow);
+    setIsPastStartDate(isStartTimeInThePast);
 
     if (!fromToken || !toToken) {
       if (!fromToken) setShowFromTokenError(true);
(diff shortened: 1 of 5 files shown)


Diogomartf:
on the `start date`
- we should change the start date to show _"Now"_ (till the user selects something) and then all the "normal" logic applies
- if the user does not change the `start date` we take it as `Date.now` (not the time in which the browser rendered but the time the confirm button is clicked)

this can be done in another PR, let me know if you have any questions

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -322,7 +329,15 @@ export const Stackbox = () => {
             </div>
             <div className="space-y-1">
               <div className="flex flex-col border divide-y md:divide-y-0 md:flex-row rounded-2xl border-surface-50 md:divide-x divide-surface-50">
-                <div className="flex flex-col w-full p-3 space-y-2 hover:bg-surface-25">
+                <div
+                  className={cx(
+                    "flex flex-col w-full p-3 space-y-2 hover:bg-surface-25",
+                    {
+                      "border border-danger-500 rounded-l-2xl":
```
Diogomartf, line 336: Small detail: I would change the danger color to 200, to not be so strong. 
```suggestion
                      "border border-danger-200 rounded-l-2xl":
```

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -331,7 +346,15 @@ export const Stackbox = () => {
                     className="w-full"
                   />
                 </div>
-                <div className="flex flex-col w-full p-3 space-y-2 hover:bg-surface-25">
+                <div
+                  className={cx(
+                    "flex flex-col w-full p-3 space-y-2 hover:bg-surface-25",
+                    {
+                      "!border !border-danger-500 !rounded-r-2xl":
```
Diogomartf, line 353: sane as above

```suggestion
                      "!border !border-danger-200 !rounded-r-2xl":
```

berteotti (dismissed):
Looks good to me!

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -89,14 +88,15 @@ export const Stackbox = () => {
     FREQUENCY_OPTIONS.hour
   );
   const [startDateTime, setStartDateTime] = useState<Date>(
-    new Date(startDateTimeTimestamp)
+    new Date(Date.now())
```
Diogomartf, line 91: Instead of `Date.now()`, let's put it as null.

And then bellow we on `openConfirmStack` we can check if the user didn't change the input. If it didn't change, we set the `Date.now()` there so it will be the time it's submited.

```js
const startDate = startDateTime ? startDateTime.getTime() : Date.now().getTime();
```

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -112,11 +112,19 @@ export const Stackbox = () => {
   };
 
   const openConfirmStack = () => {
-    const endTimeBeforeStartTime =
-      endDateTime.getTime() <= startDateTime.getTime();
-    const tokenAmountIsZero = tokenAmount === "0";
+    const startDate = startDateTime.getTime();
```
Diogomartf, line 115:
```suggestion
const startDate = startDateTime ? startDateTime.getTime() : Date.now().getTime();
```

packages/app/components/stackbox/Stackbox.tsx
berteotti: We need this value so it can be displayed in the date time input. Value is a required prop

Diogomartf: can we do `Date.now()` there?

packages/app/components/stackbox/Stackbox.tsx
Diogomartf: can't we do Date.now() there?

Diogomartf (approved):

berteotti:
Can you add some logic to prevent the user from selecting an hour in the past? If the user saves an hour in the past, we set it to current time. In this way, we always show "Now" even if the user selects a date in the past.

PR #84, opened 2023-09-04
feat: add landing

Diogomartf:
- install next js, tailwind
- add base page copy from app:
- ui components
- public images
- tailwind config

image


packages/landing/constants/index.ts
```diff
@@ -0,0 +1 @@
+export const STACKLY_APP_URL = "https://stackly-ui.pages.dev/"; // "https://stackly.eth.limo/";
```
berteotti, line 1: This should be a ENV variable, right?

packages/landing/app/layout.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export default function RootLayout({
+  children,
+}: {
+  children: React.ReactNode;
+}) {
+  return (
+    <html lang="en">
+      <body className="font-sans bg-surface-25 text-em-high">
+        <Navbar />
```
berteotti, line 22: Navbar can't be part of UI, it will be different in the landing page

packages/landing/next.config.js
```diff
@@ -0,0 +1,23 @@
+/** @type {import('next').NextConfig} */
+const nextConfig = {
+  output: "export",
+  distDir: "../../dist",
```
berteotti, line 4: `output` can be the default one, as it will be a SSR app, `distDir` must be inside the package, different from the App `distDir`

berteotti:
We need a way to have 2 previews. 1 pointing to app and one pointing to landing page

packages/landing/app/layout.tsx
Diogomartf: yeah, I'll move it to `components`

packages/landing/components/index.ts
```diff
@@ -0,0 +1,3 @@
+import { Navbar } from "./navbar/Navbar";
+
+export { Navbar };
```
berteotti, line 3: We don't need this. My suggestion works in the same way.

```suggestion
export * from "./navbar";
```

packages/landing/constants/index.ts
```diff
@@ -0,0 +1,4 @@
+export const STACKLY_LANDING_URL =
+  process.env.STACKLY_LANDING_URL ?? "https://stackly.app/";
+export const STACKLY_APP_URL =
+  process.env.STACKLY_APP_URL ?? "https://stackly-ui.pages.dev/"; // "https://stackly.eth.limo/";
```
berteotti, line 4: No need for the comment here. And we can point to the correct default URL. `STACKLY_APP_URL` can point to `https://stackly-ui.pages.dev/`.

packages/landing/components/index.ts
Diogomartf: Ok!

packages/landing/constants/index.ts
berteotti: And what about updating the default value? `"https://stackly-ui.pages.dev/"` to `"https://stackly.eth.limo/"`

Diogomartf: fixed.

berteotti (approved):
Good job Diogo!

PR #85, opened 2023-09-05
Landing: Add steps section

Diogomartf:
image


berteotti:
Mobile UI doesn't look good. Font sizes and margin should be different in mobile.

packages/landing/app/page.tsx
```diff
@@ -5,54 +5,101 @@ import { STACKLY_APP_URL } from "@/constants";
...
+      <section className="max-w-6xl px-6 mx-auto">
+        <div className="mt-20 md:mt-32 md:mb-28">
+          <HeadingText size={4}>Using Stackly is super easy.</HeadingText>
+          <HeadingText weight="regular" className="text-em-med">
+            Create a stack in 3 steps. Create and cancel anytime.
+          </HeadingText>
+        </div>
+        <StepSection
+          step={1}
+          description="Create a stack in 3 steps.Create and cancel anytime."
```
berteotti, line 67:
```suggestion
          description="Simply choose the token you want to swap from and then choose the token you want to stack."
```

packages/landing/app/page.tsx
```diff
@@ -5,54 +5,101 @@ import { STACKLY_APP_URL } from "@/constants";
...
+              Total transactions:
+            </BodyText>
+            <BodyText size={3} weight="bold">
+              1232
+            </BodyText>
+          </div>
+        </div>
+      </section>
       <div className="h-[1px] bg-gray-100 mt-10"></div>
+      <section className="max-w-6xl px-6 mx-auto">
```
berteotti, line 58: Margin top should be in the section, right? The section wants to define the spacing from the sibling sections. Also, I think this should be padding, as we are going to have sections with different background colors and the spacing must have the same background color as the section. Also using max-w and mx-auto will prevent the background from filling the entire space. I think sections should be 100% width, without margin and with some background color, and a child div to define spacing and margins. Maybe in this way we don't need to define a margin to the section.

berteotti: In this particular section, the color is fine but for the hero it's not. Other sections will have different backgrounds too.

packages/landing/app/page.tsx
```diff
@@ -5,54 +5,101 @@ import { STACKLY_APP_URL } from "@/constants";
...
+    <div className="max-w-md space-y-10 md:space-y-14">
+      <div className="w-fit px-5 py-2 bg-primary-100 rounded-[56px] text-em-med">
+        <span className="text-black/30">Step</span> {step}/3
+      </div>
+      <HeadingText weight="regular" size={2} className="text-em-med">
+        {description}
+      </HeadingText>
+    </div>
+    <Image
+      className="mt-10"
```
berteotti, line 98: From the design, the image doesn't have margin, it's the step section the one with margin.

packages/landing/app/page.tsx
```diff
@@ -5,54 +5,101 @@ import { STACKLY_APP_URL } from "@/constants";
...
+  step: number;
+  description: string;
+}
+
+const StepSection = ({ step, description }: StepProps) => (
+  <div className="flex flex-col justify-between py-12 md:py-16 md:flex-row">
+    <div className="max-w-md space-y-10 md:space-y-14">
+      <div className="w-fit px-5 py-2 bg-primary-100 rounded-[56px] text-em-med">
+        <span className="text-black/30">Step</span> {step}/3
+      </div>
```
berteotti, line 92: Font weight should be 600

berteotti:
Default font weight should be 500

packages/landing/app/page.tsx
```diff
@@ -5,54 +5,101 @@ import { STACKLY_APP_URL } from "@/constants";
...
-      <div className="mx-auto flex items-center px-5 py-2 bg-black/5 rounded-[20px] w-fit space-x-6 mt-20">
-        <div className="flex items-center space-x-2">
-          <BodyText size={3} weight="medium" className="text-em-med">
-            Total Stacks created:
-          </BodyText>
-          <BodyText size={3} weight="bold">
-            732
-          </BodyText>
+    <main className="mt-14">
+      <section className="max-w-6xl px-6 mx-auto">
```
berteotti, line 9: Background is different

packages/landing/app/page.tsx
Diogomartf: fixed

packages/landing/app/page.tsx
Diogomartf: fixed

packages/landing/app/page.tsx
Diogomartf: fixed.

packages/landing/app/page.tsx
Diogomartf: fixed

packages/landing/app/page.tsx
Diogomartf: fixed

Diogomartf:
> Mobile UI doesn't look good. Font sizes and margin should be different in mobile.

We're using regular components from `stackly ui` , it's out of scope of this PR to change them. Will do it when we have the whole landing structure.

Diogomartf:
> > Mobile UI doesn't look good. Font sizes and margin should be different in mobile.
> 
> We're using regular components from `stackly ui`  out of the box, it's out of scope of this PR to change them. Will do it when we have the whole landing structure.

berteotti (approved):

PR #89, opened 2023-09-06
update: add package script to build:landing

Diogomartf:


package.json
```diff
@@ -12,7 +12,7 @@
   ],
   "scripts": {
     "build": "turbo run build",
-    "build:app": "cd packages/sdk && yarn typechain && turbo run build --filter=@stackly/sdk --filter=app",
+    "build:app": "cd packages/sdk && yarn typechain && turbo run build --filter=@stackly/sdk --filter=app --filter=landing",
```
berteotti, line 15: Should we have a different build for the landing? We don't need to build the app to have the landing. I suggest `build:landing`

Diogomartf: fixed

berteotti (approved):

PR #100, opened 2023-09-14
fix: Stack shows it has remaining funds when it has not

Diogomartf:
**before**
image
image

**after**
image
image


ElRodrigote (approved):

berteotti (approved):
Honestly I don't know if I like this solution. How low should the `SMALL_FRACTION` be? Will approve because I can't think of anything better right now.

PR #101, opened 2023-09-14
poc: Safe AA

berteotti:
This PR is a draft of Safe AA on Stackly. Code and UX are not for review, this is just a Proof of Conecpt so we can test if and how Safe AA works with Stackly.


Diogomartf:
Tested login with email and it didn't work.

Diogomartf:
I'll not use it. Don't think a native web3 user will use it.

My question is, will people from outside of web3 be interested in using stackly currently?

We're starting to grow our socials, but it's also web3 focused, so I in the current state I don't think it's worth it.

Diogomartf:
should we close this PR?

PR #103, opened 2023-09-15
fix: safe autoconnect doesn't allow to disconnect

Diogomartf:
This PR aims to fix disconnecting issue that was introduced with auto connect.

Looks like it still works on Safe:

image

The connected wallet is the Safe wallet and not the user.


Diogomartf:
I tried a as custom app on safe and it look likes it works.

image

The connected wallet is the safe wallet and not the user wallet.

berteotti (approved):

PR #108, opened 2023-09-29
[SWA-185][FEAT] - Strategies module + "Popular Strategies" tab

ElRodrigote:
## Fixes: [SWA-185](https://linear.app/swaprhq/issue/STK-185/add-popular-strategies-strategy-tab)

# Description
* Add Sparkles icon ✨ .
* Abstract token information into constants.
* Abstract token logo sizes into a single source of truth (a single type).
* Minor modifications in `CaptionText` and `Button` components for compatibility.
* Create strategies constants, the 5th one is for test purposes. ⚠️ SHOULD BE REMOVED BEFORE MERGING ⚠️ 
* Create strategies context to handle changes across many modules.
* Create `StrategyCard` component and `Strategies` module.
* Wire up new `Strategies` module behavior into the `Stackbox` component.

# Visual evidence
https://www.loom.com/share/463abf3f963444239ad4c11717a129c7?sid=7c2766e9-b133-497f-9039-f8418a52c1bc

# How to test the changes

1) Pull this branch
2) Run the project locally
3) Go to Swapr landing page
4) Connect your wallet
5) Play around selecting and deselecting strategy cards, try updating any of the inputs from the stackbox.


packages/app/components/strategies/Strategies.tsx
```diff
@@ -0,0 +1,58 @@
...
+            >
+              {category.strategies.map((strategy: Strategy, stratIndex) => {
+                const isSelected = selectedStrategy?.id === strategy.id;
+
+                return (
+                  <StrategyCard
+                    key={stratIndex}
+                    onClick={() =>
+                      setSelectedStrategy(isSelected ? null : strategy)
+                    }
```
berteotti, line 46: This click is on the card or on the button? Should we move this logic down to the component?

packages/app/components/strategies/Strategies.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <div className="max-w-lg mx-auto mt-8">
+      <Tab.Group>
+        <Tab.List className="mb-2">
+          <div className="flex space-x-2">
+            {strategyCategories.map((category, catIndex) => (
+              <Tab
+                className={tabButtonStyles}
+                key={catIndex}
+                onClick={() => setSelectedStrategy(null)}
```
berteotti, line 25: Should we clean the strategy when clicking on the Tab?

packages/app/ui/text/CaptionText.tsx
```diff
@@ -31,6 +33,7 @@ export const CaptionText = ({
         boldStyles({ weight }),
         captionTextStyles({ size, className })
       )}
+      onClick={onClick}
```
berteotti, line 36: Text should not have `onClick` properties

packages/app/components/strategies/constants.tsx
```diff
@@ -0,0 +1,80 @@
...
+    label: () => (
+      <>
+        <Icon className="mr-1.5" name="sparkles" size={14} /> Strategies for you
+      </>
+    ),
+    strategies: [
+      
+    ],
+  },
+  */
```
berteotti, line 29: I don't think that we should have this comment here, linear task is enough I think,

berteotti (commented):
Good job Rorry, left a few comments.

packages/app/components/strategies/constants.tsx
Diogomartf: 💯

Diogomartf:
**Design feedback**

image

- We need hover effect when hovering on the card
- I think the card should be clickable and not only the "select" button.

When not connected it doesn't show the stack "preview", It would be great if it shows.

image




missing space at the bottom
image

Diogomartf:
When I click on "reset stack" I expect the stackbox to go to initial state. It just de-selected the strategy which makes little sense.

image

can we reset the values?

packages/app/components/strategies/Strategies.tsx
ElRodrigote: Right now is in the button, we can do it over the whole card. Will move the handler to the whole card.

Not sure about moving down the logic into the Card, I'd keep it here and change what triggers it.

ElRodrigote:
> **Design feedback**
> 
> image
> * We need hover effect when hovering on the card
> * I think the card should be clickable and not only the "select" button.
> 
> When not connected it doesn't show the stack "preview", It would be great if it shows.
> 
> image
> missing space at the bottom image

* About the hover I asked Akash to take a look and create a state for that
* When no connected the preview is not displaying because we're taking the from and to tokens from the tokens with balances list. To do what you suggest I should apply a workaround that checks if there's a wallet connected or not, and use one list type or the other (with all the corresponding updates). Dunno if worth it right now with our current app state. WDYT?

Diogomartf:
Maybe the preview when not connected is out of scope of this PR but I think we should plan to do that.

Good work Rorry.

berteotti:
> * When no connected the preview is not displaying because we're taking the from and to tokens from the tokens with balances list. To do what you suggest I should apply a workaround that checks if there's a wallet connected or not, and use one list type or the other (with all the corresponding updates). Dunno if worth it right now with our current app state. WDYT?

I think we should have 2 lists. Normal token list and token list with balances. Token list with balances should only be used in token list modal, the other should be used everywhere.

packages/app/components/strategies/Strategies.tsx
berteotti: I gave it a think and I find it better to move the logic to the component. Less props the better IMO.

packages/app/components/strategies/Strategies.tsx
berteotti: Sounds good to me.

packages/app/components/strategies/Strategies.tsx
berteotti: You just need to pass the strategy prop. Everything else can be assessed on the component. And even can remove `useStrategyContext` from here. What do you think?

packages/app/app/page.tsx
```diff
@@ -1,9 +1,10 @@
-import { Stackbox } from "@/components";
+import { Stackbox, Strategies } from "@/components";
 
 export default function Home() {
   return (
-    <div className="w-full mx-auto">
+    <div className="w-full py-10 mx-auto space-y-8">
```
berteotti, line 5: Can you add the same `py-10` style to stacks/page please?

ElRodrigote:
> > * When no connected the preview is not displaying because we're taking the from and to tokens from the tokens with balances list. To do what you suggest I should apply a workaround that checks if there's a wallet connected or not, and use one list type or the other (with all the corresponding updates). Dunno if worth it right now with our current app state. WDYT?
> 
> I think we should have 2 lists. Normal token list and token list with balances. Token list with balances should only be used in token list modal, the other should be used everywhere.

Yes!

Right now we have something like this (at least we have the groundwork done), the pending work is around using them consistently and doing typechecking on their use throughout the components and use cases

packages/app/components/strategies/StrategyCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+    setIsSelected(selectedStrategy?.id === strategy.id);
+  }, [selectedStrategy, strategy.id]);
+
+  return (
+    <div
+      className={cx(
+        "p-3 bg-white shadow-md min-w-60 rounded-2xl cursor-pointer",
+        {
+          "!bg-primary-900": isSelected,
+          "shadow-lg": isHovered,
```
berteotti, line 41: can this be done with css only? `hover:shadow-lg`

packages/app/components/strategies/StrategyCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+    {
+      label: "Total Amount",
+      totalAmount: `${strategy.totalSellAmount} ${sellToken.symbol}`,
+    },
+    { label: "No. of days", totalAmount: strategy.daysAmount },
+  ];
+
+  useEffect(() => {
+    setIsSelected(selectedStrategy?.id === strategy.id);
+  }, [selectedStrategy, strategy.id]);
```
berteotti, line 33: I think `const isSelected = selectedStrategy?.id === strategy.id;` works, every time `selectedStrategy` changes there is a re-render.

packages/app/components/strategies/StrategyCard.tsx
Diogomartf: Yes

Diogomartf: IMO the hover shadow effect isn't working, it's not noticeable. I would discuss with @akashtyagi027 adding a subtle change to the bg.

I think `hover:bg-surface-25` works much better.

ElRodrigote: We need to change 2 different styles on hover:
* The card shadow 
* The "select" button style

The `hover:` from Tailwind is the pseudo-selector for a specific element. If we want to target the two elements (the card + the button inside it) we need a way to achieve this other than the pseudo-selector :c

Agree that maybe instead of using `lg` shadow on hover we could use another styling to be more impactful. Will send the link to this convo to Akash c:

packages/app/components/strategies/StrategyCard.tsx
ElRodrigote: You're right lol, good catch! Thanks :D

packages/app/components/strategies/constants.tsx
```diff
@@ -0,0 +1,62 @@
...
+    label: "Popular Strategies",
+    strategies: [
+      {
+        id: 1,
+        buyToken: WETH,
+        daysAmount: 30,
+        frequency: FREQUENCY_OPTIONS.day,
+        sellAmountPerTimeframe: 100,
+        sellToken: USDC,
+        totalSellAmount: "3000",
```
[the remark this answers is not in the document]
ElRodrigote: Yes! We talked about this during a daily I think the previous week, @0xVenky will review these strategies and come up with new ones before merging :D

Thanks for noticing!

packages/app/components/strategies/StrategyCard.tsx
berteotti: I don't think we need to highlight the button when hovering the card. We only need to highlight the card with a different box shadow.

Diogomartf: Not sure if it's necessary, but It is also possible to change the `bg` on the `button` when hovering the card in `tailwind` with `group-hover` selector.

docs:
https://tailwindcss.com/docs/hover-focus-and-other-states#differentiating-nested-groups

Diogomartf:
I think clicking on "x" icon on a strategy should do the same thing as "reset stack".

packages/app/app/stacks/page.tsx
```diff
@@ -11,7 +11,7 @@ export default function Page() {
   if (isDisconnected) return <NoWalletState />;
 
   return (
-    <div className="space-y-8">
+    <div className="py-10 space-y-8">
```
Diogomartf, line 14: I don't think this spacing is good because we already doing padding top spacing on layout, we should do the spacing y there.

packages/app/app/page.tsx
Diogomartf, line 5: I think we need more space up top, otherwise it too close to the nav.

I think `py-16` is better suited.

packages/app/app/stacks/page.tsx
Diogomartf: We should converge the spacing with the home page.

It should have the same spacing top.

We probably should do this spacing on the `RootLayout` which is shared by the two pages.

```jsx
<div className="px-4 py-12 md:py-16 mx-auto md:px-0">{children}</div>
```

And remove spacing top from `YourStacksLayout`

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -257,14 +316,20 @@ export const Stackbox = () => {
...
+      })}
+    >
       <div className="py-4 border shadow-lg border-surface-50 rounded-2xl">
         <div className="flex items-end justify-between px-5 pb-4 border-b border-surface-50">
           <SelectTokenButton
             label="Deposit from"
             onClick={openTokenPicker}
             token={fromToken}
-            className={cx({ "animate-wiggle": showFromTokenError })}
+            className={cx("min-w-[32px]", {
```
Diogomartf, line 330: why add this style?

packages/app/components/strategies/StrategyCard.tsx
ElRodrigote: Whaaaaaaaaaaaat that's so cool hahaha didn't knew about these pseudo-modifies, good one Diogo! Thanks a lot :D

I'll get this into the design chat and check with Akash c:

packages/app/app/stacks/page.tsx
ElRodrigote: +1 to normalizing spacing, will do!

packages/app/components/stackbox/Stackbox.tsx
Diogomartf: I think it's ok for now.

Diogomartf (approved):

Diogomartf:
Let's add mainnet support on strategies before merging.

packages/app/contexts/StrategyContext.tsx
```diff
@@ -0,0 +1,65 @@
...
+  sellToken: Token;
+  totalSellAmount: string;
+};
+
+const throwStrategyContextError = () => {
+  throw new Error("No StrategyContext available");
+};
+
+interface StrategyContextProps {
+  selectedStrategy: Strategy | null;
```
Diogomartf, line 22: I think what we want here is for the type to handle:`undefined`

```suggestion
  selectedStrategy?: Strategy;
```

packages/app/contexts/StrategyContext.tsx
```diff
@@ -0,0 +1,65 @@
...
+  setShouldResetStackbox: React.Dispatch<React.SetStateAction<Boolean>>;
+  shouldResetStackbox: Boolean;
+}
+
+const StrategyContext = createContext<StrategyContextProps>({
+  selectedStrategy: null,
+  setSelectedStrategy: throwStrategyContextError,
+  setShouldResetStackbox: throwStrategyContextError,
+  shouldResetStackbox: false,
+});
```
Diogomartf, line 33:
```suggestion
And here we just don't define selectedStrategy. 

const StrategyContext = createContext<StrategyContextProps>({
  setSelectedStrategy: throwStrategyContextError,
  setShouldResetStackbox: throwStrategyContextError,
  shouldResetStackbox: false,
});
```

ElRodrigote:
Closed this PR to open up a new one and test new Cloudflare deployments with `bun` instead `yarn`

PR #109, opened 2023-10-06
feat: use searchParams to keep stackbox state on url

berteotti:
Sharable stacks via query params.

Screenshot 2023-10-06 at 11 08 22


Diogomartf:
I think the ability to share a stack is good!

But the button is totally out of context, how do I know it's for sharing a stack? 

I think we can get rid of the button and always put the state of the params on url and whenever someone wants to share a stack is just copying the URL which is a expected behavior.

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -256,6 +263,57 @@ export const Stackbox = () => {
...
+    if (sellToken) setFromToken(getTokenFromList(sellToken) || undefined);
+    if (buyToken) setToToken(getTokenFromList(buyToken) || undefined);
+    if (amount) setTokenAmount(amount);
+    if (startTime) setStartDateTime(new Date(parseInt(startTime, 10)));
+    if (endTime) setEndDateTime(new Date(parseInt(endTime, 10)));
+    if (interval)
+      setFrequency(
+        FREQUENCY_OPTIONS[interval as keyof typeof FREQUENCY_OPTIONS] ||
+          FREQUENCY_OPTIONS.hour
+      );
```
Diogomartf, line 314: can we make a function for this, like `setupStackbox` or something like that.

Diogomartf:
I would update the title for something like: _use searchParams to keep stackbox state on url_

Diogomartf:
is this slate?

Seems like the same PR as this: https://github.com/SwaprHQ/stackly-ui/pull/126

Let's close this.

Diogomartf:
Is this PR active?

If not please close, let's keep the PRs tab clean.

PR #110, opened 2023-10-06
[SWA-191][FEAT] - Add Mainnet support for stackly

ElRodrigote:
### Description
* Overwrite selected StartTime for the Stackbox, if the date is in the past

### How to test the changes
1) Pull this branch
2) Run the project locally
3) Go to the Stackly homepage, connect your wallet
4) Top right network selector should allow you to connect to Ethereum network
5) Try creating a stack


packages/sdk/src/vaults/constants.ts
```diff
@@ -1,32 +1,52 @@
...
+  [ChainId.ETHEREUM]: "0xc97ecbdba20c672c61e27bd657d4dfbd2328f6fa",
   [ChainId.GNOSIS]: "0x521aC6dA724aD873977d3d0b5d7B4a52E1dC52b7",
 };
 
 /**
- * COW's settlement address list
+ * COW's settlement address list, CoW Protcol GPv2VaultRelayer
  */
 export const COW_SETTLEMENT_ADDRESS_LIST: Record<ChainId, string> = {
   [ChainId.ETHEREUM]: "0xC92E8bdf79f0507f65a392b0ab4667716BFE0110",
```
berteotti, line 43:
```suggestion
  [ChainId.ETHEREUM]: "0x9008D19f58AAbD9eD0D60971565AA8510560ab41",
```

This will fix it!

berteotti (changes requested):
Found what was wrong with the stack creation. You can update it and test it :)

packages/sdk/src/vaults/constants.ts
ElRodrigote: Thanks a lot for checking it!

packages/app/components/token-picker/constants.ts
```diff
@@ -1,43 +1,16 @@
...
-  },
+import {
+  GNOSIS_GNO,
+  GNOSIS_SWAPR,
+  GNOSIS_WETH,
+  GNOSIS_WXDAI,
+  MAINNET_USDC,
+  MAINNET_WETH,
+  TokenFromTokenlist,
+} from "@/models/token";
```
Diogomartf, line 9: I think I would prefer to have something like:

```suggestion
import {
  gnosis,
  mainnet,
  TokenFromTokenlist,
} from "@/models/token";
```

packages/app/components/token-picker/constants.ts
```diff
@@ -1,43 +1,16 @@
...
-    decimals: 18,
-    logoURI: "/assets/images/tokens/swapr.png",
-    chainId: 100
-  }
-];
+export const TOKEN_PICKER_COMMON_TOKENS: {
+  [chainId: number]: TokenFromTokenlist[];
+} = {
+  1: [MAINNET_USDC, MAINNET_WETH],
+  100: [GNOSIS_GNO, GNOSIS_SWAPR, GNOSIS_WETH, GNOSIS_WXDAI],
```
Diogomartf, line 15:
```suggestion
  1: [mainnet.USDC, mainnet.WETH],
  100: [gnosis.GNO, gnosis.SWAPR, gnosis.WETH, gnosis.WXDAI],
```

packages/app/models/token/tokens.ts
Diogomartf, line 1: maybe we should have a file per chain which separated the tokens: gnosis.ts and mainnet.ts

packages/subgraph/bin/build-subgraph.ts
```diff
@@ -3,12 +3,16 @@ import { stringify as yamlStringify } from "yaml";
 
 import { config } from "./config";
 
+const SUPPORTED_NETWORKS = ["mainnet", "xdai"];
```
Diogomartf, line 6: do thegraph still use `xdai` name? 😳

ElRodrigote: https://thegraph.com/docs/en/developing/supported-networks/

You're right! I'll update that legacy CLI name, thanks for the heads-up!
image

ElRodrigote:
> The token lists are not correct on mainnet
> 
> selecting a token on mainnet
> 
> image

The token list is correct luckily but there is a misbehavior on detecting the chain swap and re-fetching the correct list, working on it right now! Thanks a lot for raising this up <3

image

Diogomartf:
This link should use the explorerLink with the correct chain funcion

image

it's using gnosis on mainnet

ElRodrigote:
> This link on Modal Header should use the `explorerLink` with the correct chain function.
> 
> image
> it's using gnosis on mainnet

Ahhh I see, I'll refactor the `txExplorerLink` function because this one is for addresses instead transactions. Thanks for pointing out!

Diogomartf (approved):

PR #116, opened 2023-10-17
feat: on stacks created show starts in X minutes our X hours 

Diogomartf:
issue: https://linear.app/swaprhq/issue/STK-178/on-stacks-created-we-should-show-starts-in-x-minutes-our-x-hours-if-is

image

Given this example, I created a stack on 20 of September and it is 20 minutes from starting that info would be better.

I would say the logic should be:
- if 0 minutes way say it's "Starting..."
- if less than 60 minutes show the minutes from order.
- if less than 24h show hours

After that show as it is "Starts on 20th September"

---
**screenshots**
image

image
image


packages/app/components/StacksTable.tsx
```diff
@@ -151,23 +151,60 @@ const CellWrapper = ({ children }: PropsWithChildren) => (
...
+    )}`;
+  } else if (hoursUntilStart < 24) {
+    dateMessage = `Starts in about ${hoursUntilStart} ${pluralize(
+      hoursUntilStart,
+      "hour"
+    )}`;
+  } else {
+    dateMessage = `Starts on ${formatTimestampToDateWithSuffix(
+      stackOrder.orderSlots[0]
+    )}`;
```
berteotti, line 206: What about using `formatDistance` from  "date-fns" here?

Diogomartf: Are we using, `date-fns` lib? If so I'll update it, if not, I rather have this code.

berteotti (approved):
Looks awesome to me!

ElRodrigote (approved):

PR #120, opened 2023-10-19
[SWA-203][PERF] - Improve RPC usage when using multicall

ElRodrigote:
## Fixes: [SWA-203](https://linear.app/swaprhq/issue/STK-203/improve-rpc-usage-when-using-multicall)

# Description
* Wrap batch fetch balances inside a `useCallback()` hook 
* WAGMI `configureChain` batch flag added

# How to test the changes

1) Pull this branch
2) Run the project locally
3) Go to Swapr landing page
4) Connect your wallet
5) Try to open the token picker for any connected chain


packages/app/contexts/TokenListContext.tsx
```diff
@@ -1,199 +1,206 @@
...
+        .call({
+          to: MULTICALL_ADDRESS,
+          data: multicallInterface.encodeFunctionData("aggregate", [callArray]),
+        })
+        .then((response: any) => {
+          const callResult = multicallInterface.decodeFunctionResult(
+            "aggregate",
+            response
+          );
+          const resultados = callResult.returnData;
```
berteotti, line 126: Let's keep it in english

berteotti: Maybe we can call it `returnData``

packages/app/providers/wagmi-config.ts
```diff
@@ -4,14 +4,14 @@ import { gnosis, mainnet } from "wagmi/chains";
...
+import { RPC_LIST } from "@/constants";
 
 const chainJsonRpc: Record<number, { http: string }> = {
   [ChainId.GNOSIS]: {
-    http: RPC_GNOSIS,
+    http: RPC_LIST[100],
   },
   [ChainId.ETHEREUM]: {
-    http: RPC_MAINNET,
+    http: RPC_LIST[1],
```
berteotti, line 14: WeE must use `chainId` here

berteotti (commented):
Good job Rorry, just left a few comments

packages/app/providers/wagmi-config.ts
ElRodrigote: I don't fully understand why or how we should use chainId here if we're setting up the RPC URL for each chain.

Could you please elaborate a bit? :c

packages/app/contexts/TokenListContext.tsx
ElRodrigote: Oh lol I haven't checked that I forgot to rename some variables, thanks for catch that!

packages/app/providers/wagmi-config.ts
Diogomartf: I think it's just in case chain id changes for GNOSIS or ETHEREUM.  Also it's more explicit in the code if you don't know what does the values 1 or 100 mean.

By using `RPC_LIST[ChainId.GNOSIS]` we don't have to care about that.

berteotti: Instead of using 1 let's user `ChainId.ETHEREUM`, this way we use the constant. Same for Gnosis.

packages/app/constants/index.ts
```diff
@@ -1,10 +1,9 @@
 // RPC endpoints
 
-export const RPC_GNOSIS =
-  process.env.RPC_GNOSIS ?? "https://rpc.gnosis.gateway.fm";
-
-export const RPC_MAINNET =
-  process.env.RPC_MAINNET ?? "https://eth.meowrpc.com/";
+export const RPC_LIST: { [chainId: number]: string } = {
+  1: process.env.RPC_MAINNET ?? "https://eth.meowrpc.com/",
+  100: process.env.RPC_GNOSIS ?? "https://rpc.gnosis.gateway.fm",
```
berteotti, line 5: Let's do the same here. Use ChainId constants.

packages/app/providers/wagmi-config.ts
ElRodrigote: Ohhh you mean something like `ChainId.Gnosis` for example? 
If that's so, I get it now, fully agree

Diogomartf (approved):

PR #126, opened 2023-10-25
feat: search params state

berteotti:
Add search params to Stackly so we can share and save stacks


packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -224,9 +321,39 @@ export const Stackbox = () => {
...
+      openConfirmStack();
+    }
+  }, [
+    isStackConfirmed,
+    isRefetchingBalance,
+    isTokenListLoading,
+    fromToken.symbol,
+    balance?.symbol,
+    openConfirmStack,
+    isModalOpen,
```
Diogomartf, line 355: what's the idea with this hook?

Diogomartf:
**this PR breaks strategies** 

When clicking on strategies, the tokens does not change.

berteotti:
> **this PR breaks strategies 🚨**
> 
> When clicking on strategies, the tokens does not change.

image


Working here. How did you get that behaviour?

packages/app/components/stackbox/Stackbox.tsx
berteotti: This hook is for opening the confirm stack when `stackConfirmed` query param is true. I honestly don't think we need this for now but it was suggested and wanted to try it out.

Diogomartf:
> > **this PR breaks strategies 🚨**
> > When clicking on strategies, the tokens does not change.
> 
> image
> 
> Working here. How did you get that behaviour?

if you click a strategy the tokens does not go to the searchParams.

I just get this url:

https://feat-search-params-state.stackly-ui.pages.dev/?tokenAmount=800&frequency=week&startDate=1698420102209&endDate=1703262102209

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -122,24 +173,50 @@ export const Stackbox = () => {
...
-  }, [chain]);
+  useEffect(() => {
+    if (
+      chainId &&
+      switchNetwork &&
+      chainId !== chain?.id &&
+      Boolean(ChainId[chainId])
+    ) {
+      switchNetwork(chainId);
+    }
```
ElRodrigote, line 217: **MINOR**, non-blocking:
No need to check for `chainId` if we're going to use it in a logic comparison, as it would default into a falsy value if it's `undefined`, and then the other checks would handle the setter execution
```suggestion
if (chainId !== chain?.id && Boolean(ChainId[chainId]) && switchNetwork) {
      switchNetwork(chainId);
    }
```

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -83,35 +98,71 @@ const balanceOptions = [
...
+        : DEFAULT_TOKENS_BY_CHAIN[ChainId.GNOSIS].to
+    )
+  );
+  const [tokenAmount, setTokenAmount] = useQueryState(
+    "tokenAmount",
+    parseAsString.withDefault("")
   );
-  const [startDateTime, setStartDateTime] = useState<Date>(
-    new Date(Date.now())
+  const [frequency, setFrequency] = useQueryState(
```
ElRodrigote, line 148: **MINOR**, non-blocking:
Dunno why but the `useQueryState()` is not inferring correctly the type for `frequency`
```suggestion
  const [frequency, setFrequency]: [
    FREQUENCY_OPTIONS,
    (frequency: FREQUENCY_OPTIONS) => {}
  ] = useQueryState(
```

ElRodrigote (dismissed):

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -122,24 +173,50 @@ export const Stackbox = () => {
...
+      setFromToken(getTokenFromList(fromTokenParam));
+    }
+
+    const toTokenParam = searchParams.get("toToken");
+    if (
+      toTokenParam &&
+      toToken.address.toLowerCase() != toTokenParam.toLowerCase()
+    ) {
+      setToToken(getTokenFromList(toTokenParam));
+    }
```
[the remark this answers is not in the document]
berteotti: I do this because it's easier to read. It's just a "style" choice, but if you declare it inside an if it gets indented and feels worse to read.

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -122,24 +173,50 @@ export const Stackbox = () => {
...
-  }, [frequency]);
+  }, [isTokenListLoading]);
 
-  // set default tokens
   useEffect(() => {
-    const chainId = chain?.unsupported
-      ? ChainId.GNOSIS
-      : chain?.id ?? ChainId.GNOSIS;
+    if (chain) setChainId(chain.id);
+  }, [chain, chain?.id, setChainId]);
```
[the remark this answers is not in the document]
berteotti: This were linter suggestion, but you are right

packages/app/components/stackbox/Stackbox.tsx
berteotti: Typescript doesn't allow this because `chainId` type is `number | null` and we can't check an enum with null.

packages/app/components/stackbox/Stackbox.tsx
berteotti: It ins't? On my side it is

Screenshot 2023-10-31 at 11 31 56

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -83,35 +98,71 @@ const balanceOptions = [
...
+    onSuccess(data) {
+      setChainId(data.id);
+    },
+  });
+  const { address } = useAccount();
+
+  const TokenParser = createParser({
+    parse: (address: string) => getTokenFromList(address),
+    serialize: (token) => token?.address || "",
   });
```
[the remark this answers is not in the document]
berteotti: Thanks, good suggestion!

packages/app/components/stackbox/Stackbox.tsx
Diogomartf: Yes, I suggested and it's cool but not worth it right now. Does not work when no wallet connected and has other issues like increasing the complexity of the code.

Let's remove it.

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -80,39 +94,85 @@ const balanceOptions = [
   { name: "Max", divider: BalanceDivider.MAX },
 ];
 
+const customParseAsTimestamp = createParser({
+  parse: (v) => {
+    const ms = parseInt(v);
+    if (Number.isNaN(ms)) {
+      return null;
+    }
```
Diogomartf, line 102: what happens if it returns null? 

Shouldn't it return now if doesn't have start time?

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -80,39 +94,85 @@ const balanceOptions = [
   { name: "Max", divider: BalanceDivider.MAX },
 ];
 
+const customParseAsTimestamp = createParser({
```
Diogomartf, line 97: this function feels out place :/

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -80,39 +94,85 @@ const balanceOptions = [
...
+  parse: (v) => {
+    const ms = parseInt(v);
+    if (Number.isNaN(ms)) {
+      return null;
+    }
+
+    const paramDate = new Date(ms);
```
Diogomartf, line 104: Not sure why but I see it as plural.

```suggestion
    const paramsDate = new Date(ms);
```

packages/app/components/stackbox/Stackbox.tsx
berteotti: If it returns `null` it will use the default value. This was copied from the library so our custom `parseAsTimestamp`

packages/app/components/stackbox/Stackbox.tsx
berteotti: Where should we define it? It's only being used here so I feel it should be used here. Open to suggestions

packages/app/components/stackbox/Stackbox.tsx
ElRodrigote: When I was checking the code locally it started to highlight a few places where we were using `frequency` in that file, maybe it was some cached stuff

packages/app/components/stackbox/Stackbox.tsx
Diogomartf: it's very related to `searchParams` so the only place I see fit is creating a `searchParams.ts` on utils folder.

But if you prefer I'm okay with letting it live here for now.

packages/app/components/stackbox/Stackbox.tsx
ElRodrigote: I'm starting to guess we have diff Typescript and Typechecks running.
When I checked this locally it worked w/o warnings even in console

ElRodrigote (dismissed):
LGTM 🚀

packages/app/components/stackbox/Stackbox.tsx
berteotti: Let's leave it here for now. If we need to use it on another file I would move it then.

Diogomartf (approved):

PR #127, opened 2023-10-26
[SWA-199][INFRA] - Track click events on Strategies

ElRodrigote:
## Fixes: [SWA-199](https://linear.app/swaprhq/issue/STK-199/track-click-events-on-strategies)

# Description
* Prune unused files and code
* Update env vars and re-arrange constants, types and contexts where they belong
* Update FA existing context and apply an event name to Strategy Cards

# Visual Evidence
https://app.usefathom.com/?range=last_30_days&site=XJTHZCQS -> this is the link to Stackly dapp (dev env) site analytics

Screenshot: 
image

# How to test the changes

1) Pull this branch
2) Run the project locally
3) Go to Stackly landing page
4) Connect your wallet
5) Clikcing a Strategy Card should trigger a FA event


packages/app/app/layout.tsx
```diff
@@ -42,13 +45,22 @@ export const metadata: Metadata = {
...
   return (
     <html lang="en" className={stabilGrotesk.variable}>
       <body className="font-sans bg-fixed bg-surface-25 bg-auto-100 bg-matrix-and-green-gradient text-em-high">
+        <Script
+          src="https://cdn.usefathom.com/script.js"
+          data-auto
+          data-site={siteId}
+          data-spa="hash"
+          defer
+        />
```
Diogomartf, line 60: Having this script here seems odd.

packages/app/components/strategies/StrategyCard.tsx
```diff
@@ -39,6 +41,7 @@ export const StrategyCard = ({ strategy }: StrategyCardProps) => {
       onClick={() => {
         setSelectedStrategy(isSelected ? null : strategy);
         setShouldResetStackbox(isSelected);
+        trackClick(EVENTS.CLICK.STRATEGY_CARD);
```
Diogomartf, line 44: We should also track which strategy was clicked.

packages/app/contexts/FathomAnalyticsContext.tsx
Diogomartf, line 1: I think the file name should be consistent with the name of the context, in this case: `AnalyticsContext`.

packages/app/contexts/FathomAnalyticsContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+  useEffect(() => {
+    if (!siteId) throw new Error("Site ID not set, skipping Fathom analytics");
+  }, []);
+
+  const analyticsContextValues = useMemo(
+    () => ({
+      trackClick: (eventName: EventName, opts?: EventOptions) => {
+        window.fathom.trackEvent(eventName, opts);
+        console.log(
+          `Click event emitted: ${eventName}.${opts ? " Options: " : ""}`,
```
Diogomartf, line 36: I don't think we should log here.

packages/app/constants/events.ts
Diogomartf, line 1: Having mix feeling about moving this logic to constants, feels very distant from where we have analytics logic.

Also events name it's only related to analytics, so It's better to call this file `analytics.ts`, so we know it's constants related to the analytics.

But honestly I would prefer to have a file named `constants.ts` in the `analytics` folder. It's closer. This is not business related logic, so I would prefer to contain everything in one place or the bare minimum.

packages/app/app/layout.tsx
Diogomartf: Not sure if preferred the `loadFathom` code.  Do you know if works the same?

ElRodrigote: All the third-party scripts we load (at least in Nextjs apps) should be preferred to be done with the `<Script />` component from Nextjs unless we have some other way to load/config them (like Tailwind).

The component also provides a few other props like "onLoad" or "onError" to trigger stuff as needed. Same goes for HTML `<head />` vs Nextjs `<Head />`.

These are optimized components to work with next, provided by them. 
Besides it's easier to read and identify IMHO, and simplifies the code as we load FA in a single place and do not use a JS function to manually pick, create, and modify a script tag. We were doing this because we don't have an actual HTML file to update on Next.

packages/app/components/strategies/StrategyCard.tsx
ElRodrigote: I was thinking about that, which would be a nice way to track it?

Maybe using the token pair? Something like this maybe?
```ts
        trackClick(
          `${EVENTS.CLICK.STRATEGY_CARD} ${strategy.buyToken}/${strategy.sellToken}`
        );
```

WDYT?

packages/app/contexts/FathomAnalyticsContext.tsx
ElRodrigote: It's just a util function to know which event are we triggering but I can remove it, honestly, I don't have strong feelings about it.
Removing it now :D

packages/app/constants/events.ts
ElRodrigote: I felt it a bit odd to, but wanted to check your thoughts about this, so it's kinda a confirmation now. Thanks for that :D

I'd like to:
* Keep the `packages\app\constants\urls.ts` file 
* And as you suggest, rename and move `packages\app\constants\events.ts` -> `packages\app\analytics\constants.ts`

WDYT?

Diogomartf: YES. That would work.

The `urls.ts` was good decision.

packages/app/components/strategies/StrategyCard.tsx
Diogomartf: I think we can put it all, since there are strategies with the same pairs.

`${EVENTS.CLICK.STRATEGY_CARD}-${strategy.buyToken}/${strategy.sellToken}-${stategy.frequency}-${stategy.amount}`

Putting it all allows the person viewing the stats to know everything about the strategy easily.

But if you prefer just `amount` or `frequency` is enough to distinguish.

packages/app/app/layout.tsx
Diogomartf: Thanks for the explanation. I think we can go with this.

packages/app/components/strategies/StrategyCard.tsx
ElRodrigote: I'll take this into Stackly channel so we can bring Venky into the discussion as well, so we're all on the same page :D (because I don't have strong opinions around this TBH)

Diogomartf: I think we can decide here tbh, just go to wtv you feel like.

Diogomartf: if we need more info we change it.

Diogomartf: Let's go with the complete one:

`${EVENTS.CLICK.STRATEGY_CARD}-${strategy.buyToken}/${strategy.sellToken}-${stategy.frequency}-${stategy.amount}`

We have the data there so let's expose it. Seeing that event makes it clear the strategy without any calc.

ElRodrigote: I used this formula + conditioning the click event trigger only when the clicked strategy is NOT selected.

packages/app/analytics/constants.ts
```diff
@@ -0,0 +1,7 @@
+export type EventName = string;
+
+export const EVENTS = {
+  CLICK: {
+    STRATEGY_CARD: "click/strategy-card",
```
Diogomartf, line 5: Wondering if with the new `cardClickEventName` function, we still need this constant file. 

Because is just being used there. I think we can add directly there.

What do you think?

ElRodrigote: I did it this way just in case the future we create more click events or other event categories (like internal/external link events to our social platforms or form submissions that are not clicked)

Diogomartf: Okay we can keep it for now.

Diogomartf:
A screenshot of the events on useFathom and link would be nice.👌

ElRodrigote:
> A screenshot of the events on useFathom and link would be nice.👌

https://app.usefathom.com/?range=last_30_days&site=XJTHZCQS -> this is the link to Stackly dapp (dev env) site analytics

Screenshot: 
image

Updating the PR description now

Diogomartf (approved):

PR #128, opened 2023-10-27
Update: Change words from "worth" -> "swapping" on preview stack

Diogomartf:
On the preview stack I think "worth" is confusing. People asked me what it meant.

After discussing we decided to go with "swapping" as it tells what we do.

**now**
image

**before**
image


ElRodrigote (approved):

berteotti (approved):

berteotti (changes requested):
Is it using the best words in this case?

berteotti (approved):
LGTM

PR #129, opened 2023-10-31
update: show wrapped tokens instead of native tokens on account

Diogomartf:
Update token balance to show wrapped version, since we don't allow using native tokens.

**Now**
image

image

### DIFF
--- a/packages/app/components/ConnectButton.tsx
+++ b/packages/app/components/ConnectButton.tsx
@@ -25,7 +25,7 @@ const CustomConnectButton = ({
   });
 
   const TOKEN_BY_CHAIN: { [chainId: number]: string } = {
-    [ChainId.ETHEREUM]: WETH[1].address,
+    [ChainId.ETHEREUM]: WETH[ChainId.ETHEREUM].address,
     [ChainId.GNOSIS]: WXDAI.address,
   };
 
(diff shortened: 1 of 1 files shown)


packages/app/components/ConnectButton.tsx
```diff
@@ -17,13 +18,24 @@ const CustomConnectButton = ({
...
   size: SizeProps;
 }) => {
+  const { chain } = useNetwork();
   const { data: avatar } = useEnsAvatar({
     name: ensName,
     chainId: 1,
   });
+
+  const TOKEN_BY_CHAIN: { [chainId: number]: string } = {
+    [ChainId.ETHEREUM]: WETH[1].address,
```
ElRodrigote, line 28: **MINOR**, non-blocking:

WDYT to abstract that `1` into a const to get rid of the magic number?

ElRodrigote (dismissed):

packages/app/components/ConnectButton.tsx
Diogomartf: yeah, true, we can use `ChainId.ETHEREUM`.

ElRodrigote (dismissed):

berteotti (approved):

PR #130, opened 2023-10-31
[SWA-198][INFRA] - Track click events on Landing page "Launch App" buttons

ElRodrigote:
## Fixes: [SWA-198](https://linear.app/swaprhq/issue/STK-198/add-analytics-to-stackly-landing-page)

# Description
* Create Anaytics context, FA installation script. and `.env.example` file
* Update `ButtonLink` component and types to use `onClick` handlers
* Apply "Launch App" click tracking + site traffic analytics to the Landing page

# Visual Evidence
https://app.usefathom.com/?range=last_30_days&site=CRHSMTWK -> this is the link to Stackly dapp (dev env) site analytics

Screenshot: 
image

# How to test the changes

1) Pull this branch
2) Run the project locally
3) Go to Stackly Landing 
4) Connect your wallet
5) Clicking any "Launch App" or "Stack Now" button should trigger a FA event


packages/landing/analytics/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const EVENTS = {
+  CLICK: {
+    LAUNCH_APP: "click/launch-app",
```
Diogomartf, line 3: Since we have this file I think we can have all the different events here instead of defining them on the spot.

`LAUNCH_APP_NAVBAR_DESKTOP`, `LAUNCH_APP_NAVBAR_MOBILE`, `LAUNCH_APP_HERO`, etc

ElRodrigote: Oh maybe, I like that approach.

My train of thought was around "ok the event is `launch app` and we have different places where we trigger it"

Both works for me to be honest

Diogomartf: Since we have the file I think we can move here.

Diogomartf:
I think we're missing this button on the CTA.

Diogomartf:
what do you think of tracking clicked Qs on the FAQ? Is that too much?

image

Since we have the setup I'm wondering if we should go a bit further.

ElRodrigote:
> what do you think of tracking clicked Qs on the FAQ? Is that too much?
> 
> image
> Since we have the setup I'm wondering if we should go a bit further. Might give us ideas to what people have more questions.
> 
> Also the links on navbar, we could also check if they we are being clicked: "faqs" and "how it works".

Added missing event tracking + created FAQ tracking

berteotti (approved):
LGTM! Good job Rorry

PR #135, opened 2023-11-08
[SWA-206][FEAT] - Reset stackbox on network switch or stack creation

ElRodrigote:
## Fixes: [SWA-210](https://linear.app/swaprhq/issue/STK-210/update-pipelines-and-build-process)

# Description
* Update `bun dev` commands per project (dApp and landing)
* Reset the stackbox on network switch or stack creation

# Visual Evidence 
https://www.loom.com/share/676905c320fe4b589458ecc91bb51e83?sid=9f247128-9179-41ce-a5cf-299c21b2892e

# How to test the changes
1) Pull this branch
2) Run the project locally
3) Go to Stackly dApp
4) Connect your wallet
5) Play around selecting strategies and then switching the network, or try creating a stack and wait for it to be successfully created. Any of these actions should reset the stackbox.


packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -227,20 +228,32 @@ export const Stackbox = () => {
...
       setFrequency(FREQUENCY_OPTIONS.hour);
       setStartDateTime(new Date(Date.now()));
       setEndDateTime(new Date(endDateByFrequency[frequency]));
     };
 
+    if (!isTokenListLoading && shouldResetStackbox) {
+      resetDefaultFormValues();
+      setShouldResetStackbox(false);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Diogomartf, line 249: why do we need to disable this?

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -683,6 +695,7 @@ export const Stackbox = () => {
           onSuccess={() => {
             closeModal(ModalId.CONFIRM_STACK);
             openModal(ModalId.SUCCESS_STACK_TOAST);
+            setShouldResetStackbox(true);
```
Diogomartf, line 698: can't we just call a function here, like`resetStackbox()` ?

I find it odd to change the state and have hook that catches that and reset it. Seems like it's more complicated.

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -227,20 +228,32 @@ export const Stackbox = () => {
...
+        chain?.id && !chain?.unsupported ? chain.id : ChainId.GNOSIS;
+
+      deselectStrategy();
+      setFromToken(DEFAULT_TOKENS_BY_CHAIN[chainId].from);
+      setToToken(DEFAULT_TOKENS_BY_CHAIN[chainId].to);
+      setTokenAmount("0.0");
       setFrequency(FREQUENCY_OPTIONS.hour);
       setStartDateTime(new Date(Date.now()));
       setEndDateTime(new Date(endDateByFrequency[frequency]));
     };
```
Diogomartf, line 243: Can't we move this function out of the hook, and then just call it when we need it?

ElRodrigote: I rather have the context to handle the conditional trigger and the hook listening to the condition update, as this condition also updates in the `StrategyCard` component.

package.json
```diff
@@ -15,7 +15,8 @@
     "build": "cd packages/sdk && bun typechain && cd ../.. && turbo run build lint",
     "build:app": "cd packages/sdk && bun typechain && turbo run build --filter=@stackly/sdk --filter=app",
     "build:landing": "cd packages/sdk && bun typechain && turbo run build --filter=@stackly/sdk --filter=landing",
-    "dev": "turbo run dev",
+    "dev:app": "cd packages/app && turbo run dev",
+    "dev:landing": "cd packages/landing && turbo run dev",
```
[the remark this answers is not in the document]
Diogomartf: Yes, I think we can keep it for now.

packages/app/components/stackbox/Stackbox.tsx
Diogomartf: Not sure, @berteotti what do you think?

packages/app/components/stackbox/Stackbox.tsx
Diogomartf: I think we should call `resetDefaultFormValues()` on those occasions we need to reset the stackbox.
That function shouldn't be tied to an hook, it can be in a hook, but it can also be used here for eg.

For me it makes sense that `onSuccess` I reset the stackbox, so I call `resetDefaultFormValues()` but here we are updating a boolean state that will trigger a `resetDefaultFormValues` somewhere.  

I think this is straight forward behavior, by hiding this complexity on a hook, we're making more difficult to reason on code and be more error prone.

(we should rename `resetDefaultFormValues` to `resetStackbox` or `resetStackboxForm`)

berteotti: Agree with Diogo here, we don't need to listen for a boolean to trigger a reset. Maybe the best solution is to create a Stack values provider, where we provide a `resetStackValues()`.

Diogomartf: why do we need a provider to do that?

Can't we just have it here `resetStackBox()`, and reset when we need?

Diogomartf: I also don't think `shouldResetStackbox` makes sense in the strategy provider, it's not related to strategy, it's related to the stackbox.

packages/app/components/stackbox/Stackbox.tsx
ElRodrigote: As we spoke offline, I'll create a new context to handle the Stackbox form values separately and consume that context both in the `StrategyCard` (to reset the form values when needed)  and the `Stackbox` component.

I'll resolve this comment thread for now c:

packages/app/components/stackbox/Stackbox.tsx
ElRodrigote: As we spoke offline, I'll create a new context to handle the Stackbox form values separately and consume that context both in the `StrategyCard` (to reset the form values when needed)  and the `Stackbox` component.

I'll resolve this comment thread for now c:

Diogomartf:
can you merge with develop? Showing a big diff.

berteotti:
> can you merge with develop? Showing a big diff.

Yeah, a rebase will help

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -134,48 +97,18 @@ export const Stackbox = () => {
...
-    parseAsTimestamp.withDefault(new Date(endDateByFrequency[frequency]))
-  );
-  const [chainId, setChainId] = useQueryState("chainId", parseAsInteger);
+  const [fromToken, setFromToken] = stackboxFormState.fromTokenState;
+  const [toToken, setToToken] = stackboxFormState.toTokenState;
+  const [tokenAmount, setTokenAmount] = stackboxFormState.tokenAmountState;
+  const [frequency, setFrequency] = stackboxFormState.frequencyState;
+  const [startDateTime, setStartDateTime] = stackboxFormState.startDateState;
+  const [endDateTime, setEndDateTime] = stackboxFormState.endDateState;
+  const [chainId, handleChainIdChange] = stackboxFormState.chainIdState;
```
Diogomartf, line 111: To be consistent, let's use `setChainId` instead of handleChainIdChange
```suggestion
  const [chainId, setChainID] = stackboxFormState.chainIdState;
```

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -134,48 +97,18 @@ export const Stackbox = () => {
   const { chain } = useNetwork();
   const { switchNetwork } = useSwitchNetwork({
     onSuccess(data) {
-      setChainId(data.id);
+      handleChainIdChange(data.id);
```
Diogomartf, line 100: is there a reason to change from `setChainId`, it's seems more consistent with the rest of things.

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -625,7 +546,7 @@ export const Stackbox = () => {
                   className="text-primary-800"
                   onClick={() => {
                     deselectStrategy();
-                    setShouldResetStackbox(true);
+                    resetDefaultFormValues();
```
Diogomartf, line 549: much nicer :)

packages/app/components/strategies/StrategyCard.tsx
```diff
@@ -15,8 +19,8 @@ interface StrategyCardProps {
 }
 
 export const StrategyCard = ({ strategy }: StrategyCardProps) => {
-  const { selectedStrategy, setSelectedStrategy, setShouldResetStackbox } =
-    useStrategyContext();
+  const { resetDefaultFormValues } = useStackboxFormContext();
```
Diogomartf, line 22: I think we can name it `resetFormValues` , by reseting I assume it goes to default. 

```suggestion
  const { resetFormValues } = useStackboxFormContext();
```

packages/app/contexts/StackboxFormContext.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+const throwStackboxFormContextError = () => {
+  throw new Error("No StackboxFormContext available");
+};
+
+/**
+ * @TODO enhance `stackboxFormState` typing
+ * @see https://github.com/youha-info/next-query-state#background-on-types
+ * @see https://www.typescriptlang.org/docs/handbook/2/generics.html
+ */
```
Diogomartf, line 58: why this comments? 

Honestly I don't like to leave comments like this. If it's important let's do it now, if not  I would delete the comment.

packages/app/contexts/StackboxFormContext.tsx
```diff
@@ -0,0 +1,178 @@
...
+interface StackboxFormContextProps {
+  resetDefaultFormValues: () => void;
+  stackboxFormState: any;
+}
+
+const StackboxFormContext = createContext<StackboxFormContextProps>({
+  resetDefaultFormValues: throwStackboxFormContextError,
```
Diogomartf, line 65: I would rename it to `resetFormValues`, I think reset is enough to intent that it goes to reset state.

```suggestion
  resetFormValues: throwStackboxFormContextError,
```

packages/app/contexts/StackboxFormContext.tsx
Diogomartf: creating a task on linear is fine tho

packages/app/components/stackbox/Stackbox.tsx
ElRodrigote: As defined in an offline chat, will rename this function into `setChainId` as it would be an expected behavior to reset the form values when changing the chain.

packages/app/components/stackbox/Stackbox.tsx
ElRodrigote: As defined in an offline chat, will rename this function into `setChainId` as it would be an expected behavior to reset the form values when changing the chain.

packages/app/contexts/StackboxFormContext.tsx
ElRodrigote: Totally agree, left the comment to signal a place I wanted attention in the PR. After an offline talk, we decided to remove the comment and handle this task in a separate ticket.

Diogomartf (approved):

PR #139, opened 2023-11-27
[STK-209][FEAT] - Enable select network without a connected wallet

ElRodrigote:
## Fixes: [STK-209](https://linear.app/swaprhq/issue/STK-209/enable-select-network-without-connected-wallet)

# Description
* Moved some code out from the Stackbox and into SelectNetwork (concern separation)
* Created custom code to handle the Chain switch as app state 
* Updated the way we consume the chain switch where it's needed

# Visual evidence
image
image

# How to test the changes

1) Pull this branch
2) Run the project locally
3) Go to Stackly dApp
4) Connect your wallet
5) You should be able to change the selected chain either by connecting the wallet or with your wallet disconnected from the dApp. The Stackbox, Strategies, and TokenPicker should display relevant information either way.

### DIFF
--- a/packages/app/components/ConnectButton.tsx
+++ b/packages/app/components/ConnectButton.tsx
@@ -1,11 +1,12 @@
 "use client";
 
+import { ChainId, WETH, WXDAI } from "@stackly/sdk";
 import { ConnectKitButton } from "connectkit";
-import { useEnsAvatar, useBalance, useNetwork } from "wagmi";
 import Image from "next/image";
+import { useAccount, useBalance, useEnsAvatar, useNetwork } from "wagmi";
+
 import { BodyText, Button, SizeProps } from "@/ui";
 import { useAutoConnect } from "@/hooks";
-import { ChainId, WETH, WXDAI } from "@stackly/sdk";
 
 const CustomConnectButton = ({
   address,
@@ -19,9 +20,10 @@ const CustomConnectButton = ({
   size: SizeProps;
 }) => {
   const { chain } = useNetwork();
+  const { isConnected } = useAccount();
   const { data: avatar } = useEnsAvatar({
     name: ensName,
-    chainId: 1,
+    chainId: ChainId.ETHEREUM,
   });
 
   const TOKEN_BY_CHAIN: { [chainId: number]: string } = {
@@ -31,9 +33,10 @@ const CustomConnectButton = ({
 
   const { data: balance } = useBalance({
     address: address,
-    token: chain
-      ? (TOKEN_BY_CHAIN[chain.id] as `0x${string}`)
-      : (TOKEN_BY_CHAIN[ChainId.GNOSIS] as `0x${string}`),
+    token:
+      chain && isConnected
+        ? (TOKEN_BY_CHAIN[chain.id] as `0x${string}`)
+        : (TOKEN_BY_CHAIN[ChainId.GNOSIS] as `0x${string}`),
   });
 
   const truncatedAddress = (size: number) =>
--- a/packages/app/components/ConnectButton.tsx
+++ b/packages/app/components/ConnectButton.tsx
@@ -3,10 +3,11 @@
 import { ChainId, WETH, WXDAI } from "@stackly/sdk";
 import { ConnectKitButton } from "connectkit";
 import Image from "next/image";
-import { useAccount, useBalance, useEnsAvatar, useNetwork } from "wagmi";
+import { useAccount, useBalance, useEnsAvatar } from "wagmi";
 
 import { BodyText, Button, SizeProps } from "@/ui";
 import { useAutoConnect } from "@/hooks";
+import { useNetworkContext } from "@/contexts";
 
 const CustomConnectButton = ({
   address,
@@ -19,7 +20,7 @@ const CustomConnectButton = ({
   ensName?: string;
   size: SizeProps;
 }) => {
-  const { chain } = useNetwork();
+  const { selectedChain } = useNetworkContext();
   const { isConnected } = useAccount();
   const { data: avatar } = useEnsAvatar({
     name: ensName,
@@ -34,8 +35,8 @@ const CustomConnectButton = ({
   const { data: balance } = useBalance({
     address: address,
     token:
-      chain && isConnected
-        ? (TOKEN_BY_CHAIN[chain.id] as `0x${string}`)
+      selectedChain && isConnected
+        ? (TOKEN_BY_CHAIN[selectedChain.id] as `0x${string}`)
         : (TOKEN_BY_CHAIN[ChainId.GNOSIS] as `0x${string}`),
   });
 
--- a/packages/app/components/SelectNetwork.tsx
+++ b/packages/app/components/SelectNetwork.tsx
@@ -1,29 +1,85 @@
 "use client";
 
-import { Button, Icon } from "@/ui";
-import { useAccount, useNetwork, useSwitchNetwork } from "wagmi";
+import { Fragment, useEffect, useState } from "react";
+
+import type { Chain } from "viem/chains";
 import { ChainIcon } from "connectkit";
 import { Listbox, Transition } from "@headlessui/react";
-import { Fragment } from "react";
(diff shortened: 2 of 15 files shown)


Diogomartf:
The network is not working when sharing a stack.

This link should open on mainnet (it has `chainId=1`).

https://7da85fee.stackly-ui.pages.dev/?tokenAmount=0.45&frequency=day&startDate=1701162851886&endDate=1706346851886&chainId=1&toToken=0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599&fromToken=0xae78736cd615f374d3085123a210448e74fc6393

packages/app/components/SelectNetwork.tsx
```diff
@@ -1,29 +1,85 @@
...
   });
-  const { chain, chains } = useNetwork();
+  const { chain } = useNetwork();
   const { isConnected } = useAccount();
+  const [selectedChain, setSelectedChain] = useState<WagmiChain>();
+  const [allowedChains, setAllowedChains] = useState<WagmiChain[] | undefined>(
+    []
+  );
+
+  const handleOfflineNetworkSwitch = (newChainId: number) => {
```
Diogomartf, line 33: I don't think offline represents not connected. I would rename it.

Eg. `handleDisconnectedNetworkSwitch` or `handleNetworkSwitchWhenNotConnected`

```suggestion
  const handleDisconnectedNetworkSwitch = (newChainId: number) => {
```

packages/app/components/SelectNetwork.tsx
```diff
@@ -1,29 +1,85 @@
...
+    config.setState((oldState: any) => {
+      const { publicClient } = oldState;
+      const { chains } = publicClient;
+
+      const offlineChain = chains?.find(
```
Diogomartf, line 38: Same here, IMO offline doesn't make sense.
In this case we're selecting a new chain if it's on allowedChain.

```suggestion
      const newChain = chains?.find(
```

Diogomartf: What if the `newChain` is not on allowed chains? Don't know if this can happen, but if take url params, maybe it can.

packages/app/components/strategies/Strategies.tsx
```diff
@@ -1,18 +1,16 @@
...
 import { tabButtonStyles } from "@/components";
 
 import { StrategyCard } from "./StrategyCard";
 import { STRATEGY_CATEGORIES } from "./constants";
 
 export const Strategies = () => {
-  const { chain } = useNetwork();
-  const chainId = chain?.id ?? ChainId.GNOSIS;
+  const { stackboxFormState } = useStackboxFormContext();
+  const [chainId] = stackboxFormState.chainIdState;
```
Diogomartf, line 13: I'm wondering if this makes sense. Why are we storing the chainId on `stackboxFormState` ?

packages/app/components/SelectNetwork.tsx
ElRodrigote: This value in specific happens `onChange` in the switch listbox, and the switch listbox only displays the allowed chains (which are the ones set in WAGMI config).

The issue you mention can happen on the Stackbox form state, will check it right now :D

packages/app/components/strategies/Strategies.tsx
ElRodrigote: We need a single source of truth for the ChainID, and we don't have a ChainContext, that's why I decided to place it in the Stackbox Form context.

packages/app/components/SelectNetwork.tsx
```diff
@@ -1,29 +1,85 @@
...
       setChainId(data.id);
     },
   });
-  const { chain, chains } = useNetwork();
+  const { chain } = useNetwork();
   const { isConnected } = useAccount();
+  const [selectedChain, setSelectedChain] = useState<WagmiChain>();
+  const [allowedChains, setAllowedChains] = useState<WagmiChain[] | undefined>(
+    []
+  );
```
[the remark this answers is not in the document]
ElRodrigote: `useNetwork` only returns valid values if a wallet is connected, if it isn't, then it returns "empty" values.

That's why I'm recurring to the `getPublicClient` from the WAGMI's `config` object
If the wallet is disconnected, `useNetwork` returns:
* `[]` for `chains`
* `undefined` for `chain`

berteotti:
After talking with @ElRodrigote , we agreed on implementing a new context in this PR. The context should store `chainId` and do most of the logic done inside `SelectNetwork`.

packages/app/components/ConnectButton.tsx
```diff
@@ -19,9 +20,10 @@ const CustomConnectButton = ({
   size: SizeProps;
 }) => {
   const { chain } = useNetwork();
+  const { isConnected } = useAccount();
   const { data: avatar } = useEnsAvatar({
     name: ensName,
-    chainId: 1,
+    chainId: ChainId.ETHEREUM,
```
Diogomartf, line 26: Nice 😎

packages/app/providers/wagmi-config.ts
```diff
@@ -1,9 +1,12 @@
+"use client";
```
Diogomartf, line 1: why do we need "use client" on this file?

packages/app/contexts/NetworkContext.tsx
```diff
@@ -0,0 +1,149 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  useEffect(() => {
+    const { chains } = config.getPublicClient();
+    const isValidChainId = getIsValidChainId(chainId);
+
+    if (!isValidChainId) setChainId(ChainId.GNOSIS);
+    if (chains) {
+      const validChainId = isValidChainId ? chainId : ChainId.GNOSIS;
```
Diogomartf, line 83: I think we are checking twice for `isValidChainId`.

packages/app/contexts/NetworkContext.tsx
```diff
@@ -0,0 +1,149 @@
...
+      const validChain = chains.find((chain) => chain.id === validChainId);
+
+      setSelectedChain(validChain ?? gnosis);
```
Diogomartf, line 86: I think we have redundant checks

packages/app/providers/wagmi-config.ts
ElRodrigote: Because I wanted to export `config` from this file and to be able to export it from the module's index.

packages/app/components/ConnectButton.tsx
```diff
@@ -19,9 +20,10 @@ const CustomConnectButton = ({
   size: SizeProps;
 }) => {
   const { chain } = useNetwork();
```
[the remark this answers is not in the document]
ElRodrigote: Not 100% sure, will check it now! Most likely we can

packages/app/components/SelectNetwork.tsx
```diff
@@ -1,29 +1,25 @@
...
+import { Listbox, Transition } from "@headlessui/react";
 
-  if (!isConnected || !chain) return <></>;
+import { Button, Icon } from "@/ui";
+import { useNetworkContext, useStackboxFormContext } from "@/contexts";
 
-  const onValueChange = (networkId: string) =>
-    switchNetwork && switchNetwork(Number(networkId));
+export const SelectNetwork = () => {
+  const { allowedChains, changeNetwork, selectedChain } = useNetworkContext();
```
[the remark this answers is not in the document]
ElRodrigote: We don't have an `unallowed` chains array but a `supported` one.

I'll rename it for simplicity's sake tho :D

packages/app/components/ConnectButton.tsx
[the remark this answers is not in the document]
berteotti: What the status on this? Lost track if we can change it or not :)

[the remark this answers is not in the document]
ElRodrigote: I totally forgot about this, but I already pushed it!

Diogomartf (approved):

PR #142, opened 2023-12-01
[STK-218][FIX] - Stack orders should show the correct cow order status

ElRodrigote:
## Fixes: [STK-218](https://linear.app/swaprhq/issue/STK-218/ui-should-show-if-it-is-a-cow-order-has-expired)

# Description
* Update the way we handle how to visualize Cow order statuses.

# Visual evidence
image

# How to test the changes
**PRE-CONDITIONS:** You need to have existing stacks with executed orders in place to test this fix

1) Pull this branch
2) Run the project locally
3) Go to Stackly dApp
4) Connect your wallet
5) Go to "Your Stacks", choose one of them, and click "View Details"
6) When the Order Details modal opens up, the right-most column should have the correct order status

### DIFF
--- a/packages/app/components/stack-modal/StackOrdersTable.tsx
+++ b/packages/app/components/stack-modal/StackOrdersTable.tsx
@@ -1,9 +1,14 @@
 import { useState } from "react";
 
 import Link from "next/link";
-import { Order as CowOrder } from "@cowprotocol/cow-sdk";
+import { Order as CowOrder, OrderStatus } from "@cowprotocol/cow-sdk";
 
-import { addressShortner, convertedAmount, formatDate } from "@/utils";
+import {
+  addressShortner,
+  capitalize,
+  convertedAmount,
+  formatDate,
+} from "@/utils";
 import {
   BodyText,
   Icon,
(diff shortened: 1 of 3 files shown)


packages/app/utils/strings.ts
```diff
@@ -0,0 +1,2 @@
+export const capitalize = (word: string) =>
+  `${word.charAt(0).toUpperCase()}${word.slice(1)}`;
```
Diogomartf, line 2: We can do this with css, tailwind has a `capitalize` className.

ElRodrigote: Lol nice, I didn't knew there was a way to do it with CSS, thanks!

packages/app/components/stack-modal/StackOrdersTable.tsx
```diff
@@ -88,6 +96,31 @@ const TableCowBody = ({
 }) => {
   const chainId = 100; // @todo use context or useNetwork hook
 
+  const renderOrderStatus = (orderStatus: OrderStatus, price?: number) => {
+    switch (orderStatus) {
+      case "fulfilled":
```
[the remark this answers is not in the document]
ElRodrigote: I tried it but was receiving a build error:
```
node:buffer
Module build failed: UnhandledSchemeError: Reading from "node:buffer" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
```

That's why I dropped it at the time. BUUUUUUUUUUT after some digging, it seems it's an issue with the version for `node-fetch` dev dependency.

After downgrading its version to `^2.6.1` I had no issues.

For more context: https://github.com/vercel/next.js/discussions/33982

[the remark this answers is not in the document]
ElRodrigote: As we talked offline with @berteotti , I'll undo this downgrade and create a local ENUM instead.

Diogomartf (approved):

packages/app/utils/index.ts
```diff
@@ -3,5 +3,6 @@ export * from "./constants";
 export * from "./datetime";
 export * from "./ethers";
 export * from "./numbers";
+export * from "./strings";
```
Diogomartf, line 6: we can remove this

```suggestion
```

packages/app/utils/strings.ts
Diogomartf: I think we can delete this file

berteotti (dismissed):
Looks good to me after fixing what Diogo mentioned.

berteotti (approved):

PR #149, opened 2024-01-17
Fixes: Switching network not working properly

Diogomartf:
This PR was mainly refactoring NetworkProvider to fixe this issue: Trying to switch to Ethereum network not working.

https://gyazo.com/c1139b2d7775f7866004f86835016575

This breaks using Stackly on Safe wallet, cause on safe to access wallets is on network base. So to access on ethereum mainnet, you can't switch to gnosis and then back to Ethereum.

### DIFF
--- a/packages/app/app/stacks/stacksOrders.tsx
+++ b/packages/app/app/stacks/stacksOrders.tsx
@@ -47,9 +47,8 @@ export const StackOrders = ({ chainId, address }: StackOrdersProps) => {
   const fetchStacks = useCallback(() => {
     getOrders(chainId, address.toLowerCase())
       .then(async (orders) => {
-        if (!orders || orders.length === 0) {
-          setCurrentStackOrders([]);
-        } else {
+        if (!orders || orders.length === 0) setCurrentStackOrders([]);
+        else {
           const stackOrders = await getStackOrders(chainId, orders);
           if (stackOrders.length > 0) setCurrentStackOrders(stackOrders);
         }
--- a/packages/app/app/stacks/stacksOrders.tsx
+++ b/packages/app/app/stacks/stacksOrders.tsx
@@ -47,10 +47,12 @@ export const StackOrders = ({ chainId, address }: StackOrdersProps) => {
   const fetchStacks = useCallback(() => {
     getOrders(chainId, address.toLowerCase())
       .then(async (orders) => {
-        if (!orders || orders.length === 0) return;
-
-        const stackOrders = await getStackOrders(chainId, orders);
-        if (stackOrders.length > 0) setCurrentStackOrders(stackOrders);
+        if (!orders || orders.length === 0) {
+          setCurrentStackOrders([]);
+        } else {
+          const stackOrders = await getStackOrders(chainId, orders);
+          if (stackOrders.length > 0) setCurrentStackOrders(stackOrders);
+        }
       })
       .finally(() => setLoading(false));
   }, [address, chainId]);
--- a/packages/app/contexts/NetworkContext.tsx
+++ b/packages/app/contexts/NetworkContext.tsx
@@ -8,7 +8,6 @@ import {
   useMemo,
   useState,
 } from "react";
-
 import type { Chain } from "viem/chains";
 import { ChainId } from "@stackly/sdk";
 import { gnosis } from "wagmi/chains";
@@ -16,7 +15,7 @@ import { useAccount, useNetwork, useSwitchNetwork } from "wagmi";
 
 import { config } from "@/providers/wagmi-config";
 import { parseAsInteger, useQueryState } from "next-usequerystate";
-import { getIsValidChainId } from "@/utils";
+import { checkIsValidChainId } from "@/utils";
 
 interface WagmiChain extends Chain {
   unsupported?: boolean;
@@ -28,14 +27,14 @@ const throwNetworkContextError = () => {
 
 interface NetworkContextProps {
   chains?: WagmiChain[];
-  chainId: ChainId;
-  changeNetwork: (newChainId: string) => void;
+  chainId?: ChainId;
+  changeNetwork: (newChainId: number) => void;
   selectedChain: WagmiChain;
-  setChainId: React.Dispatch<React.SetStateAction<ChainId | null>>;
+  setChainId: React.Dispatch<React.SetStateAction<ChainId | undefined>>;
 }
 
 const NetworkContext = createContext<NetworkContextProps>({
-  chainId: ChainId.GNOSIS,
+  chainId: gnosis.id,
   changeNetwork: throwNetworkContextError,
   selectedChain: gnosis,
   setChainId: throwNetworkContextError,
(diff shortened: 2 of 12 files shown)


packages/app/contexts/NetworkContext.tsx
```diff
@@ -51,56 +50,58 @@ export const NetworkContextProvider = ({
...
+    if (isConnected && chain) {
+      setSelectedChain(chain);
+      setSelectedChainId(chain?.id);
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [chainId, selectedChain]);
-
-  useEffect(() => {
-    if (isConnected && chain) setSelectedChain(chain);
   }, [chain, isConnected]);
```
ElRodrigote, line 105: Just thinking out loud:
This hook will trigger with similar conditions to the one above it, and they both kinda set the same values (except for the first condition that switch network)

I'm not sure if this won't bring trouble in the future

packages/app/app/stacks/stacksOrders.tsx
```diff
@@ -47,10 +47,12 @@ export const StackOrders = ({ chainId, address }: StackOrdersProps) => {
...
-        if (!orders || orders.length === 0) return;
-
-        const stackOrders = await getStackOrders(chainId, orders);
-        if (stackOrders.length > 0) setCurrentStackOrders(stackOrders);
+        if (!orders || orders.length === 0) {
+          setCurrentStackOrders([]);
+        } else {
+          const stackOrders = await getStackOrders(chainId, orders);
+          if (stackOrders.length > 0) setCurrentStackOrders(stackOrders);
+        }
```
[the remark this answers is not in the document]
Diogomartf: updated, thanks!

packages/app/contexts/NetworkContext.tsx
Diogomartf: yeah, also not so confident we need this hook, will take a second look.

Diogomartf: Ok, got it, we need this so anytime the chain changes it updates local provided state variables. As in the app we use NetworkProvider instead of wagmi `useNetwork` hook, so we need to update those whenever, chain changes. This is only needed when wallet is connected.

ElRodrigote (approved):

packages/app/contexts/TokenListContext.tsx
```diff
@@ -163,23 +166,45 @@ export const TokenListProvider = ({ children }: PropsWithChildren) => {
...
+                tokenListNoDuplicates.push(token);
+              }
+            });
+
+            const tokenListOfCurrentChainId = tokenListNoDuplicates.filter(
+              (token: TokenFromTokenlist) => token.chainId === chainId
+            );
+
+            return tokenListOfCurrentChainId;
+          };
```
berteotti, line 185: I think we don't need to wrap it inside a function as it's only create a bit more complexity on reading this code. Passing `tokenListOfCurrentChainId` to `setTokenList` will create the same effect

packages/app/contexts/TokenListContext.tsx
```diff
@@ -163,23 +166,45 @@ export const TokenListProvider = ({ children }: PropsWithChildren) => {
...
           );
+          const tokenUniqueAddresses: string[] = [];
+          const tokenListNoDuplicates: TokenFromTokenlist[] = [];
+
+          const tokenList = () => {
+            mergedTokenlistTokens.forEach((token) => {
+              if (!tokenUniqueAddresses.includes(token.address)) {
+                tokenUniqueAddresses.push(token.address);
+                tokenListNoDuplicates.push(token);
+              }
```
berteotti, line 177: We can already filter inside this condition, right? That will make this code faster since we aren't looping twice through the same array

berteotti (commented):
Made a couple of improvements suggestions. Otherwise looks good to me

packages/app/contexts/TokenListContext.tsx
ElRodrigote: We're iterating over `mergedTokenlistTokens` here, we're not iterating twice the same array.

BUT I'm 100% agree that we add a small boolean check to know if we should push a value inside the array instead of filtering it afterward

packages/app/providers/Providers.tsx
```diff
@@ -17,7 +17,7 @@ import {
 export const Providers = ({ children }: PropsWithChildren) => {
   return (
     <WagmiConfig config={config}>
-      <ConnectKitProvider mode="light">
+      <ConnectKitProvider mode="light" options={{ initialChainId: 0 }}>
```
ElRodrigote, line 20: Why `0` as initial chain ID?

Diogomartf: good question!

we do that so if we are on ethereum network on our wallet when connecting to stackly it keeps ethereum. Previously, it was triggering the wallet to change to gnosis and then again to change back to ethereum.

image

https://docs.family.co/connectkit/api-reference#initialChainId

packages/app/contexts/TokenListContext.tsx
```diff
@@ -144,42 +129,58 @@ export const TokenListProvider = ({ children }: PropsWithChildren) => {
...
+
+          const tokenUniqueAddresses: string[] = [];
+          const tokenListNoDuplicates: TokenFromTokenlist[] = [];
+
+          tokenListOfCurrentChainId.forEach((token) => {
+            if (!tokenUniqueAddresses.includes(token.address)) {
+              tokenUniqueAddresses.push(token.address);
+              tokenListNoDuplicates.push(token);
+            }
+          });
```
berteotti, line 161: I see we are filtering and then running a forEach. Can we filter inside the forEach? This way we don't need more than 1 loop through the token list.

berteotti: Also, maybe something like this

```
const tokenUniqueAddressesSet = new Set<string>();
const tokenListNoDuplicates = tokenListOfCurrentChainId.filter((token) => {
  if (!tokenUniqueAddressesSet.has(token.address)) {
    tokenUniqueAddressesSet.add(token.address);
    return true;
  }
  return false;
});
```

and filter per chain too. This might be a cleaner solution

packages/app/providers/Providers.tsx
ElRodrigote: Ahhhhhhh gotcha gotcha, makes a lot of sense now

ElRodrigote (approved):

packages/app/contexts/TokenListContext.tsx
```diff
@@ -144,42 +129,56 @@ export const TokenListProvider = ({ children }: PropsWithChildren) => {
...
 
   useEffect(() => {
+    // abort if still fetching other chain async
+    abortControllerRef.current.abort();
+    abortControllerRef.current = new AbortController();
     setupTokenList();
-  }, [address, chainId, setupTokenList]);
+
+    return () => {
+      // Cleanup: Abort ongoing fetch when component unmounts or chainId changes
```
berteotti, line 177: I think this Cleanup comment isn't accurate. This doesn't trigger when chainId changes

berteotti (approved):
Just left a comment

packages/app/contexts/TokenListContext.tsx
Diogomartf: I'll remove it

PR #160, opened 2024-02-22
check if cow order is fulfilled to calculate avg price

Diogomartf:
Fixes:
image


packages/app/models/stack-order/stack-order.ts
```diff
@@ -16,6 +17,8 @@ export const calculateStackAveragePrice = (order: StackOrder) => {
 
   order.cowOrders.forEach((cowOrder) => {
     if (cowOrder.executedBuyAmount === "0") return;
+    // @ts-expect-error
+    if (cowOrder.status !== CowOrderStatus.FULFILLED) return;
```
ElRodrigote, line 21: Our `CowOrderStatus` and CoW SDK `OrderStatus` have the same values, but TS tries to infer the types from two different enums and has collisions. IN THIS CASE, we can drop our `CowOrderStatus` and simply use the one from CoW SDK to fix this issue

```suggestion
    if (cowOrder.status !== OrderStatus.FULFILLED) return;
```

ElRodrigote (approved):

packages/app/models/stack-order/stack-order.ts
Diogomartf: Yeah, it should be that, that was my first thought. But using `OrderStatus` breaks compilation unexpectedly.
 
image

Diogomartf: Was able to use `OrderStatus` 🥳

berteotti (approved):
Awesome, much cleaner!

berteotti:
Can you rebase the branch ? Looks like build is failling too, probably rebase can solve this

PR #172, opened 2024-05-28
feat: update subgraph endpoints

berteotti:
Updates subgraph endpoints to decentralised network


ElRodrigote (approved):

packages/sdk/src/vaults/constants.ts
```diff
@@ -1,5 +1,6 @@
 import { AddressZero } from "@ethersproject/constants";
 import { ChainId } from "../constants";
+import { env } from "process";
```
Diogomartf, line 3: don't seem to be used

```suggestion
```

PR #185, opened 2024-07-05
[STK-233][FIX] -  Bug in UI order amount calcs

ElRodrigote:
## Context
After @berteotti investigation on the SC side, he found out that we were not executing a trade on the end date, so we were displaying different computation values in the UI and the SC.

## Description
* Removes the `INITIAL_ORDER` value (= `1`) from the `estimatedNumberOfOrders` calculation to mimic the current logic in the Smart Contract side.

### DIFF
--- a/packages/app/components/stackbox/ConfirmStackModal.tsx
+++ b/packages/app/components/stackbox/ConfirmStackModal.tsx
@@ -87,11 +87,9 @@ export const ConfirmStackModal = ({
   const [stackCreationTx, setStackCreationTx] = useState<Transaction>();
 
   const rawAmount = parseUnits(amount, fromToken.decimals);
-  const estimatedNumberOfOrders =
-    Math.floor(
-      (endTime.getTime() - startTime.getTime()) / frequencySeconds[frequency]
-    ) + INITAL_ORDER;
-
+  const estimatedNumberOfOrders = Math.floor(
+    (endTime.getTime() - startTime.getTime()) / frequencySeconds[frequency]
+  );
   const amountPerOrder = (parseFloat(amount) / estimatedNumberOfOrders).toFixed(
     2
   );


packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -190,6 +189,29 @@ export const Stackbox = () => {
...
+
+    oneMinuteInterval();
+
+    const id = setInterval(oneMinuteInterval, ONE_MINUTE_IN_MS);
+
+    return () => {
+      clearInterval(id);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
berteotti, line 213: Not sure if we want this. Was editing the date and it auto updated while I was editing it.

I think we should update the start time only, if we want to keep it in the present. Also I would move it to the present, and not add 1 minute.

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -484,14 +506,15 @@ export const Stackbox = () => {
                 >
                   <BodyText size={2}>Until</BodyText>
                   <DatePicker
-                    dateTime={endDateTime}
+                    dateTime={getDatePlusMinsOffset(endDateTime, ONE_MINUTE)}
```
berteotti, line 509: Why are we adding 1 minute here. The experience becomes weird.

packages/app/components/DatePicker.tsx
```diff
@@ -48,11 +53,13 @@ export function DatePicker({
   };
 
   useEffect(() => {
-    const newDate = new Date(dateTime);
+    const newDate = isEndDate
+      ? getDatePlusMinsOffset(dateTime, TEN_MINUTES)
+      : new Date(dateTime);
```
berteotti, line 58: Why are we adding this logic?

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -315,7 +335,8 @@ export const Stackbox = () => {
 
   const estimatedNumberOfOrders =
     Math.floor(
-      (endDateTime.getTime() - startDateTime.getTime()) /
+      (getDatePlusMinsOffset(endDateTime, ONE_MINUTE).getTime() -
+        startDateTime.getTime()) /
```
berteotti, line 339: Time should be added to startDateTime. Let's add 10 minutes to it.

ElRodrigote: Something like this you mean?
```ts
    Math.floor(
      (endDateTime.getTime() -
        getDatePlusMinsOffset(startDateTime, TEN_MINUTES).getTime()) /
        frequencySeconds[frequency as FREQUENCY_OPTIONS]
    ) + INITAL_ORDER;
```
    
If I do this, we'll end up having less time slots for the given frequency
So for example: 
* Start: July 21th, at 12:00
* End: July 24th, at 12:00
* Frequency: daily
* 4 WXDAI to WETH

Should display "1 WXDAI daily for 3 days" (due to the initial order)
But instead, it will display this:
Screenshot 2024-07-16 at 12 08 54

berteotti: If you create the stack, what are the stacks buying timestamps?

berteotti: We call the slots

packages/app/components/stackbox/Stackbox.tsx
Diogomartf: I think we don't need to update the UI if in the past, we just handle when submit
- if start time is on the past set it to the current time.
The confirm model will show the starting time and will allow verification

IMO this makes the code more complex without giving much upside

ElRodrigote:
Closing this PR for now, as the production code as-is displays the same information we get from the SC. In the future we will re-visit the logic we use to handle the stack slots.

PR #189, opened 2024-07-24
[STK-255][FIX] - My Stacks breaks for some one time orders

ElRodrigote:
## Fixes: 
* [STK-255](https://linear.app/swaprhq/issue/STK-255/my-stacks-breaks-for-some-one-time-orders)
* [STK-257](https://linear.app/swaprhq/issue/STK-257/order-details-display-wrong-start-and-end-dates)

## Description
* Adds different checks when we have a stack with no `orderSlots`.

## Visual Evidence
https://www.loom.com/share/6522431e773f4350af5ce0b70c93f909?sid=1fd75bc2-f862-4aa8-bf96-61e762dc12f5

### DIFF
--- a/packages/app/models/order/order.ts
+++ b/packages/app/models/order/order.ts
@@ -18,8 +18,12 @@ export const totalOrderSlotsDone = (order: Order) => {
   }, 0);
 };
 
-export const allOrderSlotsDone = (order: Order) =>
-  totalOrderSlotsDone(order) === order.orderSlots.length;
+export const allOrderSlotsDone = (order: Order) => {
+  const orderSlotsLength = order.orderSlots.length;
+  if (!orderSlotsLength) return totalOrderSlotsDone(order) === 1;
+
+  return totalOrderSlotsDone(order) === orderSlotsLength;
+};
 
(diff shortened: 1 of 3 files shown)


packages/app/models/order/order.ts
```diff
@@ -3,14 +3,27 @@ import { currentTimestampInSeconds } from "@/utils/datetime";
 import { Order } from "@stackly/sdk";
 
 export const totalOrderSlotsDone = (order: Order) => {
+  /**
+   * An order that doesn't have slots happens when we have a 1 slot order
+   * that has start and end date in a timeframe equals to or less than 60'.
+   * As it's not possible to use the slot timestamp, we use the start time
+   * for this 1-time order.
+   */
```
Diogomartf, line 11:
```suggestion
  /**
   * An order that doesn't have slots happens when we have a 1 slot order
   * that has start and end date in a timeframe equal or less than 60'.
   * As it's not possible to use the slot timestamp, we use the start time
   * for this 1 slot order.
   */
```

Diogomartf (approved):

berteotti (approved):

PR #195, opened 2024-08-06
[STK-215][FIX] - Completed stack with funds without showing "withdraw" button

ElRodrigote:
## Description
* Updates our checks for stack remaining funds, we had an edge-case bug where if the stack only spent our protocol fee and the stack completed without spending any more funds, we didn't allow the user to withdraw the remaining unspent funds.

## Visual Evidence
https://www.loom.com/share/0dbc2f9189104fa9a64e65bb2858927b?sid=fb70b53a-b4d4-4c43-868d-c0d31eb50047


berteotti (approved):
Nice jog Rorry, it's working!

image

PR #196, opened 2024-08-07
[STK-258][FIX] - Bar showing complete on the first order 

ElRodrigote:
## Description
* Updates the computations we do to set the progress bar, we were taking the executed orders as the total orders.

## Visual Evidence
Screenshot 2024-08-07 at 12 15 02


Diogomartf:
this fix misses an edge case, when a DCA does less orders than expected.

When all the funds are used we need to show complete even tho it didn't do all the expected order it did us all the funds so for the stack is complete.
 
image


the normal case seems to be working fine

image

Diogomartf (approved):

Diogomartf:
good job 🥳

berteotti (approved):

PR #199, opened 2024-08-14
[STK-256][FIX] - My Stacks not updating correctly

ElRodrigote:
## Context
This seems to be the same issue described in our previous issue for a bug in our network switch handling. For more context, please see: https://github.com/SwaprHQ/stackly-ui/pull/194

## Description
* Persists the chain ID in the search params
* Resets the stackbox values only if we're changing chains in the landing page

### DIFF
--- a/packages/app/constants/pathnames.ts
+++ b/packages/app/constants/pathnames.ts
@@ -1,0 +1,4 @@
+export const PATHNAMES = {
+  LANDING: "/",
+  STACKS: "/stacks",
+};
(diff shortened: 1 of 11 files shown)


Diogomartf:
overall working good, but I found a bug, semi related:

(showing gnosis and arbitrum on the network)
image


If you are on "my stacks" change network and click on "my stacks" it changes the network button to default network (arbitrum). It does not change the network, just the button

https://github.com/user-attachments/assets/907769a2-6dbb-4b77-b74a-92113ab1c229

packages/app/constants/pathnames.ts
```diff
@@ -0,0 +1,4 @@
+export const PATHNAMES = {
+  LANDING: "/",
```
berteotti, line 2: Let's call it `HOME` instead

Diogomartf (approved):

berteotti (approved):

PR #202, opened 2024-08-23
release v1.3.2

ElRodrigote:
* Added new token logos for CoW, Shutter, and Olas
* Added Shutter and Olas to our Gnosis token list
* Created env vars to handle our subgraph API endpoints
* Bugfixes:
  * for singe-slot and no-slot orders:
    * Make these orders cancellable under the correct scenarios
    * My stacks page was breaking the app
    * My Stacks were displaying `NaN` for average buy prices
  * Stackbox:
    * Now it hides the stack details tile if the end date is in the past.
    * Enabled the mobile users to place decimal amounts in the Amount input
    * Allowed to create stacks with a start date close to the current timestamp of the user
    * Updates the past end date error when the end date changes
  * Stack Details:
    * Now displays the correct orders progress bar.
    * Allows stack funds withdrawal for some edge cases where the stacks only spent the protocol fee
  * The Network Switch component was not picking up the right chain ID in some scenarios
  * My Stacks now updates correctly on network switch


berteotti (approved):
Good job Rorry!

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -486,6 +488,10 @@ export const Stackbox = () => {
                   <DatePicker
                     dateTime={endDateTime}
                     setDateTime={(date: Date) => {
+                      const isEndTimeBeforeStartTime =
+                        date.getTime() <= startDateTime.getTime();
+
+                      setShowPastEndDateError(isEndTimeBeforeStartTime);
```
Diogomartf, line 494: shouldn't we only set if `isEndTimeBeforeStartTime` is `true`?

ElRodrigote: If we don't set the `false` value the error won't get cleaned on a valid date change

Diogomartf (approved):
Great job on getting this release Rorry out!

It sure killed a lot of bugs.

PR #209, opened 2024-10-21
[FIX][STK-252] - Update My Stack tabs after stack cancellation

ElRodrigote:
## Fixes: [STK-252](https://linear.app/swaprhq/issue/STK-252/update-my-stack-tabs)

## Description
* Adds an optional chaining for safety navigation inside `stack-order.ts` to cover an edge case where it might break the app on a fast tab switch, to get the remaining funds of the stack.
* Re-fetches all the user orders after successfully canceling a stack

## Visual Evidence
https://www.loom.com/share/3dbdab016ac840b09b77f8e63076ced0?sid=d5e43b35-a395-4e2b-b6ae-9f3c3288924f

### DIFF
--- a/packages/app/app/stacks/stacksOrders.tsx
+++ b/packages/app/app/stacks/stacksOrders.tsx
@@ -231,6 +231,7 @@ export const StackOrders = ({ chainId, address }: StackOrdersProps) => {
                           stacks.orders,
                           stacks.sort
                         )}
+                        refetchAllOrders={fetchAllOrders}
                         refetchStacks={() =>
                           fetchStacks(index as StackStateIndex)
                         }
--- a/packages/app/components/StacksTable.tsx
+++ b/packages/app/components/StacksTable.tsx
@@ -37,6 +37,7 @@ import { ModalId, useModalContext } from "@/contexts";
 
 interface StacksTableProps {
   stackOrders: StackOrder[];
+  refetchAllOrders: () => void;
   refetchStacks: () => void;
   hasMorePages: boolean;
   hasLessPages: boolean;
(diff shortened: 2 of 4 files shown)


packages/app/app/stacks/stacksOrders.tsx
```diff
@@ -231,6 +231,7 @@ export const StackOrders = ({ chainId, address }: StackOrdersProps) => {
                           stacks.orders,
                           stacks.sort
                         )}
+                        refetchAllOrders={fetchAllOrders}
```
Diogomartf, line 234: why no use `fetchAllOrders` instead of  `refetchAllOrders`?
cause it's seems to fetchAllOrders and not necessarily has the logic for re-fetching.

berteotti: Looks like a refetch, right? Also we are using refetchStacks with the same logic

berteotti (approved):

packages/app/app/stacks/stacksOrders.tsx
ElRodrigote: Exactly as Leo suggested, the idea for this prop is that, when we successfully cancel a stack and close the modal, we re-fetch the needed data:
* The total orders count used for the tabs count (and the "All stacks count")
* The displayable orders per page

Diogomartf: but we're passing it a `fetchAllOrders` function, I would keep the same name. As the function fetches the orders. Where we use it is the re-fetch thing, but would keep the same name.

Diogomartf (approved):

packages/app/app/stacks/stacksOrders.tsx
ElRodrigote: Ah I get the semantic you suggest now, I'm ok with your suggestion, will push it in a bit.

Diogomartf (approved):

PR #212, opened 2024-11-22
[FEAT][STK-186] - Add new frequency selector

ElRodrigote:
## Fixes: [STK-186](https://linear.app/swaprhq/issue/STK-186/add-new-frequency-selector)

## Description
* Creates the new Checkbox component
* Creates the new TextInput component
* Crates a FrequencyOpitonsCard component 
* Implements everything in the Stackbox

## Visual Evidence
https://www.loom.com/share/6845f9de011f41c8b9ddb18986cd81f6?sid=207fdb56-359d-410b-a545-76d28a7e18b2


packages/app/components/stackbox/FrequencyOptionsCard.tsx
```diff
@@ -0,0 +1,134 @@
...
+interface FrequencyOptionsCardProps {
+  frequency: FREQUENCY_OPTIONS;
+  setEndDate: (date: Date) => void;
+}
+
+const defaultFrequencyOptions = {
+  [FREQUENCY_OPTIONS.hour]: ["12", "24", "72"],
+  [FREQUENCY_OPTIONS.day]: ["7", "15", "30"],
+  [FREQUENCY_OPTIONS.week]: ["5", "10", "20"],
+  [FREQUENCY_OPTIONS.month]: ["3", "6", "12"],
```
Diogomartf, line 19: here some suggestions:

```suggestion
  [FREQUENCY_OPTIONS.hour]: ["6", "12", "24"],
  [FREQUENCY_OPTIONS.day]: ["2", "5", "10"],
  [FREQUENCY_OPTIONS.week]: ["1", "2", "4"],
  [FREQUENCY_OPTIONS.month]: ["2", "6", "12"],
```
I use the day frequency, and usually is 3, 5 days. I think 7 15 30 are better reflected in weeks: 1,2,4.

packages/app/components/stackbox/FrequencyOptionsCard.tsx
```diff
@@ -0,0 +1,134 @@
...
+};
+const maxCustomFrequencies = {
+  [FREQUENCY_OPTIONS.hour]: 96,
+  [FREQUENCY_OPTIONS.day]: 365,
+  [FREQUENCY_OPTIONS.week]: 52,
+  [FREQUENCY_OPTIONS.month]: 36,
```
Diogomartf, line 25: I would say that we don't want such a big stacks (as they can be low values), so I would reduce at least the months to 24.

```suggestion
  [FREQUENCY_OPTIONS.month]: 24,
```

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -339,6 +342,8 @@ export const Stackbox = () => {
     setFromToken(toToken);
   };
 
+  console.log("frequency: ", frequency);
```
Diogomartf, line 345: log

packages/app/components/stackbox/FrequencyOptionsCard.tsx
ElRodrigote: I like it, so basically we top up to 1 year DCA orders in almost every frequency as max (unless the user wants to set them via Date pickers)

packages/app/components/stackbox/FrequencyOptionsCard.tsx
```diff
@@ -0,0 +1,134 @@
...
+    if (postiveIntegerOnly.test(newValue) && isWithinRange) {
+      setDefaultFrequency("");
+      setCustomFrequency(newValue);
+    }
+  };
+
+  useEffect(() => {
+    const placeholderText = `Custom ${frequency} (max: ${
+      maxCustomFrequencies[frequency]
+    } ${getCroppedFrequency(frequency)})`;
```
Diogomartf, line 79: image

I would remove the "d"/"w"/"m" from the placeholder. In this case would be (max: 52)

cause it might imply I have to put it, I prefer just the number.

```suggestion
    const placeholderText = `Custom ${frequency} (max: ${
      maxCustomFrequencies[frequency]
    } )`;
```

packages/app/components/stackbox/FrequencyOptionsCard.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const dateNow = new Date();
+  const endDateByDefaultFrequency = {
+    [FREQUENCY_OPTIONS.hour]: dateNow.setHours(dateNow.getHours() + timeAmount),
+    [FREQUENCY_OPTIONS.day]: dateNow.setDate(dateNow.getDate() + timeAmount),
+    [FREQUENCY_OPTIONS.week]: dateNow.setDate(
+      dateNow.getDate() + timeAmount * 7
+    ),
+    [FREQUENCY_OPTIONS.month]: dateNow.setMonth(
+      dateNow.getMonth() + timeAmount
+    ),
```
berteotti, line 42: Maybe is better to use https://date-fns.org/v4.1.0/docs/add# here. I feel like months are not being set correctly on the UI. Example: selected 12 months, but gives me +15 months

image

packages/app/components/stackbox/FrequencyOptionsCard.tsx
```diff
@@ -0,0 +1,132 @@
...
+  [FREQUENCY_OPTIONS.month]: 24,
+};
+const postiveIntegerOnly = /^$|^[0-9]+$/;
+
+const getEndDateByDefaultFrequency = (
+  frequency: FREQUENCY_OPTIONS,
+  timeAmount: number
+) => {
+  const dateNow = new Date();
+  const endDateByDefaultFrequency = {
```
berteotti, line 34: We need to have this object outside the function, and returning a function itself. Otherwise we all calculating all this different values for hour, days,..., month when we are only using 1 value. Also it creates a bug, where we are adding `timeAmount` to `dateNow`, so when using months for example, we are adding `timeAmount` to `dateNow` on the hour, day, week and months, where it should only be on months slot.

packages/app/components/stackbox/FrequencyOptionsCard.tsx
berteotti: I know see that the bug is related to this https://github.com/SwaprHQ/stackly-ui/pull/212/files#r1858630377

But we should use date-fns anyway, which as simpler syntax.

berteotti, line 42: We need to have this object outside the function, and returning a function itself. Otherwise we all calculating all this different values for hour, days,..., month when we are only using 1 value. Also it creates a bug, where we are adding `timeAmount` to `dateNow`, so when using months for example, we are adding `timeAmount` to `dateNow` on the hour, day, week and months, where it should only be on months slot.

packages/app/ui/inputs/TextInput.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export const TextInput = ({
+  className,
+  error,
+  id,
+  label,
+  onChange,
+  placeholder,
+  value,
+  wrapperClassName,
```
berteotti, line 22: We can delete this wrapperClassName for now, right? I don't see it being used

packages/app/ui/inputs/TextInput.tsx
```diff
@@ -0,0 +1,49 @@
...
+    {label && (
+      <Label htmlFor={id}>
+        <BodyText className="text-em-low">{label}</BodyText>
+      </Label>
+    )}
+    <Input
+      id={id}
+      value={value}
+      placeholder={placeholder}
+      onChange={onChange}
```
berteotti, line 35: Most of this props are already propagated on `inputProps`, right? We can not declare this if they aren't be used elsewhere.

packages/app/components/stackbox/FrequencyOptionsCard.tsx
```diff
@@ -0,0 +1,132 @@
...
+        <div className="flex gap-2">
+          {defaultFrequencyOptions[frequency].map((freqOption, idx) => {
+            const isSelected = defaultFrequency === freqOption;
+
+            return (
+              <RadioButton
+                checked={isSelected}
+                className="min-w-[70px] h-10"
+                id={freqOption}
+                key={idx}
```
berteotti, line 108: We can use the freqOption as the key, right? So, no need for idx.

packages/app/components/stackbox/FrequencyOptionsCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+    case FREQUENCY_OPTIONS.hour:
+      return add(new Date(), { hours: timeAmount });
+    case FREQUENCY_OPTIONS.day:
+      return add(new Date(), { days: timeAmount });
+    case FREQUENCY_OPTIONS.week:
+      return add(new Date(), { weeks: timeAmount });
+    case FREQUENCY_OPTIONS.month:
+      return add(new Date(), { months: timeAmount });
+    default:
+      throw new Error("Invalid frequency option");
```
berteotti, line 44: Let's not throw an error here, because we aren't catching it. Maybe a console.error is good enough in this case.

packages/app/components/stackbox/FrequencyOptionsCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+            return (
+              <RadioButton
+                checked={isSelected}
+                className="min-w-[70px] h-10"
+                id={freqOption}
+                key={freqOption}
+                name={freqOption}
+                onChange={(event) => {
+                  setDefaultFrequency(event.target.value as FREQUENCY_OPTIONS);
+                }}
```
berteotti, line 110: Maybe we should clean the text input too?

packages/app/components/stackbox/FrequencyOptionsCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+            );
+          })}
+        </div>
+        <TextInput
+          id="custom-frequency-option"
+          min={1}
+          onChange={handleCustomFrequencyChange}
+          placeholder={placeholder}
+          value={customFrequency}
+        />
```
berteotti, line 126: We are allowing 0. Maybe we shouldn't allow 0 or empty. If user has 0/empty it defaults to the frequency default value.

berteotti (approved):
Looking good to me. Only this edge case is still happening. I input 33 weeks and then change to months frequency.

image

packages/app/components/stackbox/FrequencyOptionsCard.tsx
```diff
@@ -0,0 +1,132 @@
...
+  [FREQUENCY_OPTIONS.week]: ["4", "8", "12"],
+  [FREQUENCY_OPTIONS.month]: ["2", "6", "12"],
+};
+const maxCustomFrequencies = {
+  [FREQUENCY_OPTIONS.hour]: 96,
+  [FREQUENCY_OPTIONS.day]: 365,
+  [FREQUENCY_OPTIONS.week]: 52,
+  [FREQUENCY_OPTIONS.month]: 24,
+};
+const postiveIntegerOnly = /^[1-9][0-9]*$/;
```
Diogomartf, line 28: Maybe we could add Regex, so it's super clear

```suggestion
const postiveIntegerOnlyRegex = /^[1-9][0-9]*$/;
```

packages/app/components/stackbox/FrequencyOptionsCard.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+const getEndDateByDefaultFrequency = (
```
Diogomartf, line 30: Isn't, `getDefaultEndDateFrequency`,  more clear?

```suggestion
const getDefaultEndDateFrequency = (
```

packages/app/components/stackbox/FrequencyOptionsCard.tsx
```diff
@@ -0,0 +1,132 @@
...
+  }, [frequency]);
+
+  useEffect(() => {
+    const newFrequency = customFrequency ? customFrequency : defaultFrequency;
+
+    setEndDate(getEndDateByDefaultFrequency(frequency, Number(newFrequency)));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [defaultFrequency, customFrequency]);
+
+  const placeholder = `Custom ${frequency} (max: ${maxCustomFrequencies[frequency]})`;
```
Diogomartf, line 88: Also think it makes more clear `placeholderText`
```suggestion
  const placeholderText = `Custom ${frequency} (max: ${maxCustomFrequencies[frequency]})`;
```

packages/app/components/stackbox/Stackbox.tsx
```diff
@@ -472,52 +475,66 @@ export const Stackbox = () => {
...
-                      setShowPastEndDateError(isEndTimeBeforeStartTime);
-                      deselectStrategy();
-                      setEndDateTime(date);
-                    }}
-                    timeCaption="End time"
-                    className="w-full"
-                    fromDate={startDateTime}
-                  />
+            <Checkbox
+              id="display-custom-date-range"
```
Diogomartf, line 479: I would add a `checkbox-freq` or just `checkbox` prefix to make it clear what the ID is.

```suggestion
              id="checkbox-freq-display-custom-date-range"
```

packages/app/components/stackbox/FrequencyOptionsCard.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  return (
+    <div
+      className={twMerge([
+        "flex flex-col p-3 rounded-2xl gap-1",
+        "bg-surface-25 border border-surface-50",
+      ])}
+    >
+      <BodyText className="text-em-low">Duration</BodyText>
```
Diogomartf, line 97:
```suggestion
      <BodyText className="text-em-low">Total duration of the stack</BodyText>
```

image

packages/app/components/stackbox/FrequencyOptionsCard.tsx
ElRodrigote: Agree! This single-use variable doesn't improve much more the readability

packages/app/components/stackbox/FrequencyOptionsCard.tsx
```diff
@@ -0,0 +1,132 @@
...
+                id={freqOption}
+                key={freqOption}
+                name={freqOption}
+                onChange={(event) => {
+                  setDefaultFrequency(event.target.value as FREQUENCY_OPTIONS);
+                  setCustomFrequency("");
+                }}
+                value={freqOption}
+              >
+                <BodyText className={!isSelected ? "text-em-med" : ""} size={2}>
```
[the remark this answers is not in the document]
ElRodrigote: Our current typing structure doesn't allow this suggestion, but I can use `cx` from CVA to handle it gracefully :E

THANKS!

packages/app/ui/inputs/TextInput.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <Label htmlFor={id}>
+        <BodyText className="text-em-low">{label}</BodyText>
+      </Label>
+    )}
+    <Input
+      id={id}
+      className={twMerge(
+        [
+          "border placeholder:text-sm w-full min-h-10 rounded-3xl px-3 py-2",
+          "border-surface-75 placeholder:text-em-low outline-primary-500",
```
Diogomartf, line 31: why have 2 strings? 

At least let's split it per context, I would prefer something like this:

```suggestion
          "border border-surface-75",
          "placeholder:text-sm placeholder:text-em-low ",
          "outline-primary-500   w-full min-h-10 rounded-3xl px-3 py-2"
```

Diogomartf:
image

It's seems to have a bug. When I add a number, I can delete it. I should be able to. 

For eg in the screenshot, I can't delete the 4.

packages/app/ui/inputs/TextInput.tsx
ElRodrigote: I like the suggestion, I tried to group colors on one hand and the rest on the other, but I like the context idea more.

I'll move the `outline-primary-500" color to the border context because it's kinda related.

PR #220, opened 2024-12-24
fix(STK-267): strategies not working properly with new token picker

ElRodrigote:
## Description
* Updates our existing strategies to fit the new frequency picker constrains
* Updates our frequency card picker so it can handle the strategy selection

## Visual Evidence
https://www.loom.com/share/5210e383aeb54a9891eada4b09d584fb?sid=7d350ff1-231d-41fa-8925-eb964ea87359

### DIFF
--- a/packages/app/components/stackbox/FrequencyOptionsCard.tsx
+++ b/packages/app/components/stackbox/FrequencyOptionsCard.tsx
@@ -82,7 +82,7 @@ export const FrequencyOptionsCard = ({
   );
   const [customFrequency, setCustomFrequency] = useState("");
 
-  const { selectedStrategy } = useStrategyContext();
+  const { deselectStrategy, selectedStrategy } = useStrategyContext();
 
   const handleCustomFrequencyChange = (
     event: ChangeEvent<HTMLInputElement>
@@ -103,6 +103,8 @@ export const FrequencyOptionsCard = ({
       setDefaultFrequency("");
       setCustomFrequency(newValue);
     }
+
+    if (selectedStrategy) deselectStrategy();
   };
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


packages/app/components/stackbox/FrequencyOptionsCard.tsx
```diff
@@ -2,13 +2,32 @@
...
+    case FREQUENCY_OPTIONS.hour:
+      return days * 24;
+    case FREQUENCY_OPTIONS.day:
+      return days;
+    case FREQUENCY_OPTIONS.week:
+      return daysToWeeks(days);
+    case FREQUENCY_OPTIONS.month:
+      return days / 30;
+    default:
+      throw new Error("Invalid frequency option");
```
berteotti: We are throwing an error which is never caught, maybe we should log a message and return something.

packages/app/components/strategies/constants.tsx
```diff
@@ -93,7 +93,7 @@ export const STRATEGY_CATEGORIES: { [chainId: number]: ChainStrategies } = {
         {
           id: 3,
           buyToken: gnosisTokens.WETH,
-          daysAmount: 10,
+          daysAmount: 4,
```
berteotti, line 96: I know that 10 days of hourly trades is a lot. Any specific reason for choosing 4?

ElRodrigote: It fits the current constraint of the maximum hours per hourly trade accepted in our custom timeframe input. Right now, we have a maximum of 96 hours, so 4 days * 24 hours = 96.

We can change any of these variables, but I made this change to fit the constraints explained above.

packages/app/components/stackbox/FrequencyOptionsCard.tsx
ElRodrigote: You're right! I'll console the error and return either the max or min value allowed for the selected frequency.

berteotti (approved):
Looks good!

